PR #11, opened 2022-01-09
Automatically detect HA port

felipecrs:
This simplifies the setup and prevent problems of misconfiguration.

Refs https://github.com/home-assistant/addons/pull/2307


brenner-tobias:
Great suggestion, I was not aware of the variable "core.port".
I have one doubt though: Why should we still offer to change the port at all? Also, it is technically not correct that the variable default of internal_ha_port is 8123, is it the one that is actually used by HA core.

So my suggestion is to remove the config option completely.
Do you have any thoughts?

felipecrs:
> Why should we still offer to change the port at all?

That's why I removed the option. ;)

brenner-tobias:
Oh right my bad, I thought you only made it optional at first read. Will merge it. Thanks!

felipecrs:
Thank you for the great add-on!

PR #21, opened 2022-01-19
Improve additional_hosts example

felipecrs:
- Make them consistent across the whole document
- Add an example which shows that it's possible to use a custom port

### DIFF
--- a/cloudflared/DOCS.md
+++ b/cloudflared/DOCS.md
@@ -64,10 +64,12 @@ Example extended add-on configuration:
 external_hostname: "ha.example.com"
 tunnel_name: "homeassistant"
 additional_hosts:
+  - hostname: "router.example.com"
+    service: "http://192.168.1.1"
   - hostname: "diskstation.example.com"
-    service: "http://192.168.1.5"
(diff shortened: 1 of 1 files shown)


cloudflared/DOCS.md
```diff
@@ -64,10 +64,12 @@ Example extended add-on configuration:
 external_hostname: "ha.example.com"
 tunnel_name: "homeassistant"
 additional_hosts:
+  - hostname: "router.example.com"
+    service: "http://192.168.1.1"
   - hostname: "diskstation.example.com"
-    service: "http://192.168.1.5"
-  - hostname: "website.example.com"
     service: "http://192.168.1.2"
```
brenner-tobias, line 70: I think this is great, but we could go a step further and also show that https is working. 
So my idea would be to change this to "https://192.168.1.2:5001"

brenner-tobias (changes requested):

cloudflared/DOCS.md
felipecrs: Is it good now?

brenner-tobias (approved):

PR #24, opened 2022-01-20
✨ Add option to use quick tunnels

pemontto:
# Proposed Changes

This adds a 0 config/setup option for users to create a tunnel. Not recommended for production, but provides and easy and approachable way to test Cloudflare Tunnels.

```bash
2022-01-20T09:56:33Z INF Thank you for trying Cloudflare Tunnel. Doing so, without a Cloudflare account, is a quick way to experiment and try it out. However, be aware that these account-less Tunnels have no uptime guarantee. If you intend to use Tunnels in production you should use a pre-created named tunnel by following: https://developers.cloudflare.com/cloudflare-one/connections/connect-apps
2022-01-20T09:56:33Z INF Requesting new quick Tunnel on trycloudflare.com...
2022-01-20T09:56:36Z INF +--------------------------------------------------------------------------------------------+
2022-01-20T09:56:36Z INF |  Your quick Tunnel has been created! Visit it at (it may take some time to be reachable):  |
2022-01-20T09:56:36Z INF |  https://disciplinary-tennis-breaking-under.trycloudflare.com                              |
2022-01-20T09:56:36Z INF +--------------------------------------------------------------------------------------------+
```

### DIFF
--- a/cloudflared/DOCS.md
+++ b/cloudflared/DOCS.md
@@ -17,6 +17,23 @@ Please make sure to be compliant with the
 add-on. Especially [section 2.8][cloudflare-sssa-28] could be breached when
 mainly streaming videos or other Non-HTML content.
 
+## Quick Tunnel for Testing
+
+You can get started with zero setup by using
+[Cloudflare Quick Tunnels](https://developers.cloudflare.com/cloudflare-one/connections/connect-apps/run-tunnel/trycloudflare).
+Set `quick_tunnel` to `true` , all other configuration will be ignored. After
+starting the addon check the logs for your unique randomly generated
+`trycloudflare.com` URL.
+
+Quick Tunnel add-on configuration:
+
+```yaml
+quick_tunnel: true
+external_hostname: ''
+tunnel_name: ''
+additional_hosts: []
+```
+
 ## Installation
 
 The installation of this add-on is pretty straightforward but requires some prerequisites
--- a/cloudflared/config.yaml
+++ b/cloudflared/config.yaml
@@ -27,5 +27,6 @@ schema:
       service: str
       disableChunkedEncoding: bool?
   nginxproxymanager: bool?
+  quick_tunnel: bool?
   log_level: list(trace|debug|info|notice|warning|error|fatal)?
   reset_cloudflared_files: bool?
--- a/cloudflared/rootfs/etc/cont-init.d/cloudflared-config.sh
+++ b/cloudflared/rootfs/etc/cont-init.d/cloudflared-config.sh
@@ -239,6 +239,12 @@ tunnel_uuid=""
 main() {
     bashio::log.trace "${FUNCNAME[0]}"
 
+    # Quick Tunnel with 0 config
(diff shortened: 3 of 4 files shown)


cloudflared/DOCS.md
```diff
@@ -17,6 +17,19 @@ Please make sure to be compliant with the
 add-on. Especially [section 2.8][cloudflare-sssa-28] could be breached when
 mainly streaming videos or other Non-HTML content.
 
+## Quick Tunnel for Testing
+
+You can get started with zero setup by using [Cloudflare Quick Tunnels](https://developers.cloudflare.com/cloudflare-one/connections/connect-apps/run-tunnel/trycloudflare). Set `quick_tunnel` to `true` , all other configuration will be ignored. After starting the addon check the logs for your unique randomly generated `trycloudflare.com` URL.
```
brenner-tobias, line 22: Can you shorten this so it follows the max line length?

cloudflared/config.yaml
```diff
@@ -16,12 +16,14 @@ arch:
...
   - i386
 options:
+  quick_tunnel: false
   external_hostname: ""
   tunnel_name: "homeassistant"
   additional_hosts: []
 schema:
-  external_hostname: str
-  tunnel_name: str
+  quick_tunnel: bool
```
brenner-tobias, line 24: I think we should make the `quick_tunnel` variable optional here and not define any default value in the options. This way, we can offer the option (also prominently in the Docu) without making it required for the regular user.

cloudflared/config.yaml
```diff
@@ -16,12 +16,14 @@ arch:
   - armv7
   - i386
 options:
+  quick_tunnel: false
```
brenner-tobias, line 19: See comment below: I would delete this line and make the flag optional.

brenner-tobias (changes requested):

cloudflared/config.yaml
pemontto: Agreed, have made the change

cloudflared/config.yaml
```diff
@@ -20,11 +20,12 @@ options:
   tunnel_name: "homeassistant"
   additional_hosts: []
 schema:
-  external_hostname: str
-  tunnel_name: str
+  external_hostname: str?
```
brenner-tobias, line 23: I like the suggestion to make the two variables `external_hostname` and `tunnel_name` optional as well. Nevertheless, now we do not have sufficient checks in place, in case the user starts the add-on without those two variables but also without setting `quick_tunnel`.
I think a good solution would be to check in `cloudflared-config.sh` after checking that `quick_tunnel` is zero (line 237), to also check if `external_hostname` and `tunnel_name` are set. Otherwise, we can abort with an error stating that you either have to set those two or `quick_tunnel`.
What do you think?

pemontto: Makes sense! I've made the change 👍

brenner-tobias (approved):

PR #25, opened 2022-01-21
Add originRequest support for disableChunkedEncoding flag

elcajon:
# Proposed Changes

The disableChunkedEncoding flag for origin is useful for host services using WSGI servers. 
For example Proxmox server isn't working with cloudflare tunnel without related service settings. 

See https://developers.cloudflare.com/cloudflare-one/connections/connect-apps/configuration/configuration-file/ingress#notlsverify for further information. 

## Related Issues

### DIFF
--- a/cloudflared/DOCS.md
+++ b/cloudflared/DOCS.md
@@ -70,6 +70,7 @@ additional_hosts:
     service: "https://192.168.1.2:5001"
   - hostname: "website.example.com"
     service: "http://192.168.1.3:8080"
+    disableChunkedEncoding: true
 nginxproxymanager: true
 log_level: "debug"
 ```
@@ -101,6 +102,11 @@ other systems like a diskstation, router or anything else.
 Like with the `external_hostname` of HomeAssistant, DNS entries at will be
 automatically created at Cloudflare.
 
+Add the (optional) `disableChunkedEncoding` option to a hostname, to disable
+chunked transfer encoding. This is useful if you are running a WSGI server,
+like Proxmox for example. Visit [Cloudflare Docs][disablechunkedencoding] for
+further information. 
+
 Please find below an examplary entry for two additional hosts:
 
 ```yaml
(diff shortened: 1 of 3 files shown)


elcajon:
Successfully tested (but without NPM), of course. 😉

brenner-tobias:
I think this is great and made me think. Maybe we can offer even more functionality by offering another array in the `additional_hosts` array called `originRequest`. That way, everyone can add whatever options they want and it would make adding the array to the config even easier.

What do you think?

elcajon:
> I think this is great and made me think. Maybe we can offer even more functionality by offering another array in the `additional_hosts` array called `originRequest`. That way, everyone can add whatever options they want and it would make adding the array to the config even easier.
> 
> What do you think?

I‘ve tried to while adding this one, but nesting config variables is limited to a depth of two, so I decided to use a top level variable instead. I guess we should consider adding functionality to the most common ones related to HassOS/Supervised. 
I also assume  it‘s not best practice to access options.json files directly (use bashio is preferred) - but this would result in a refactor of the init script. 

Sent from mobile, excuse typos.

brenner-tobias:
Good point. Let's start with this one then.

elcajon:
Could you publish the release in a timely manner? Feel free to contact me (or check my public repos) if you plan to set up a edge (or stable) repository with workflows.

brenner-tobias:
Done. I will also have a look at automating the release process in the next couple of days, thanks for pointing that our

brenner-tobias:
@elcajon: I just started looking into the auto-update in this repo and the [ha-addons-edge](https://github.com/brenner-tobias/ha-addons-edge).
It looks like everything is working fine in the repo, but the r[un of update in the add-on repository](https://github.com/brenner-tobias/ha-addons-edge/actions/runs/1747835669) fails. It shows `ValueError: untagged-2eea43dd07768ed394b6 is not valid SemVer string`. This untagged-2eea43dd07768ed394b6 is the same value for different runs and I have no idea what's wrong here.
Do you mind having a look?
Thanks a lot in advance.

Issue #43, opened 2022-02-10
Change of location for config and certificate files


elcajon:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

I recently thought about changing the location for cert.pem and tunnel.json from /data to /{ssl or share}/cloudflared. 
This would benefit in making both files reusable on other hosts for redundancy.

I also thought about providing the possibility to add config.yml (.yml, not .json) manually. Which, if file exists, leads to starting clouflared tunnel with the manually created config file. This would benefit allow making more complex service configuration with other ingress config options. 

In my opinion this could be realized with another optional configuration option (e.g. "data_path").

**Describe alternatives you've considered**

**Additional context**

In addition I would suggest (independent of "data_path") moving the automatically created config.json file outside the persistent /data storage as it is regenerated on every restart either way. 

@brenner-tobias  I could take a look at it tonight if you think the changes make sense as well.

brenner-tobias:
Makes sense, great suggestion. Feel free to have a go at it, happy to also look into it over the weekend.

elcajon:
#44  will add base functions and should work without interfering already existing configurations

I will add the proposed custom config.yml functionality in a separate PR.

PR #44, opened 2022-02-10
Add data_path funcationality & general maintenance

elcajon:
# Proposed Changes

- Add data_path functionality #43 
- change default config.json path from /data to /tmp #43 
- change default `external_hostname` to "null" so user can't start add-on without adding own value

### DIFF
--- a/cloudflared/DOCS.md
+++ b/cloudflared/DOCS.md
@@ -213,13 +213,13 @@ additional_hosts: []
 
 ### Option: `data_path`
 
-The `data_path` option allows to change default storage 
-location (`/data`) for the automatically created `cert.pem` and 
+The `data_path` option allows to change default storage
+location (`/data`) for the automatically created `cert.pem` and
 `tunnel.json` file.`
 
-This can be useful if you want to reuse both files on other hosts. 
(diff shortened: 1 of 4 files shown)


brenner-tobias:
Great work @elcajon, thanks a lot.
I am unsure about one thing though: We have to map the additional home assistant folders in the config files anyways (as you did). This does limit the selection of possible folders for the user which he cannot change. Now in the current config, as you correctly mentioned in the documentation, only `ssl` and `share` are possible.

1. I am wondering if it would be better to change `data_path` from a string to list(ssl|share], maybe rename it to `data_folder` or something like that and always use `${data_folder}/cloudflared` for the actual path. 
2. Additionally, it might be a good idea to also include the config folder. This one is from my perspective also looking into the next functionality the best folder to add all the cloudflared files from your computer, since you can easily access it via the file editor add-on.

elcajon:
Good idea, I'll adapt that

elcajon:
I have implemented your suggestions. I have not revised the logic of the migration for the time being, but I think the status should be sufficient for a first version.

brenner-tobias (approved):
Awesome, feel free to merge

PR #45, opened 2022-02-11
Add custom config.yml

elcajon:
# Proposed Changes

- > Add feature for using custom config.yml #43

### DIFF
--- a/cloudflared/rootfs/etc/cont-init.d/cloudflared-config.sh
+++ b/cloudflared/rootfs/etc/cont-init.d/cloudflared-config.sh
@@ -318,7 +318,7 @@ createCustomDNS() {
     bashio::log.trace "${FUNCNAME[0]}"
 
     # Check for configured additional hosts and create DNS entries for them if existing
-    for host in $(cat "${data_path}/config.yml" | yq e '.ingress[].hostname | select(. == "*")' -); do
+    for host in $( yq e '.ingress[].hostname | select(. == "*")' "${data_path}/config.yml" ); do
         bashio::log.info "Creating new DNS entry ${host}..."
         if bashio::var.is_empty "${host}" ; then
             bashio::exit.nok "'hostname' is empty, please check your config file."
--- a/cloudflared/rootfs/etc/cont-init.d/cloudflared-config.sh
+++ b/cloudflared/rootfs/etc/cont-init.d/cloudflared-config.sh
@@ -311,6 +311,45 @@ migrateFiles() {
     fi
 }
 
+# ------------------------------------------------------------------------------
+# Create cloudflare DNS entry for external hostname and additional hosts
+# ------------------------------------------------------------------------------
+createCustomDNS() {
+    bashio::log.trace "${FUNCNAME[0]}"
+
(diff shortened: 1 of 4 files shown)


cloudflared/rootfs/etc/cont-init.d/cloudflared-config.sh
```diff
@@ -354,6 +393,10 @@ main() {
         createTunnel
     fi
 
+    if hasCustomConfig ; then
```
brenner-tobias, line 396: I think we need to adapt this logic. If the login and the tunnel creation are done in our add-on and the config should be created in a custom way, there is no way for the user to know the tunnel UUID to put into the config in the first place. Do you know, what I mean?

So we have different possible situations:

1. hasCustomConfig and tunnel_file and cert are existing (all put there by user): fine, works already
2. !hasCustomConfig: fine, works already
2. hasCustomConfig and tunnel file is missing: create tunnel (and previously create cert if not there). Then the only way for it to work would be to automatically add the created tunnel UUID and the link to the credentials file in the custom_config. This would be quite nice but needs to be stated clearly in the docs.md. The other option here would be to not allow that combination and make the user add everything himself if he wants a custom config.

Does that make sense or am I missing something?

cloudflared/rootfs/etc/services.d/cloudflared/run
```diff
@@ -14,10 +14,17 @@ else
     if bashio::config.has_value 'data_folder'; then
         data_path="/$(bashio::config 'data_folder')/cloudflared"
     fi
+    # Check status file which exists when custom_config exists
+    if bashio::fs.file_exists "/tmp/custom_config"; then
```
brenner-tobias, line 18: We don't need that if we go with the additional config flag. Otherwise I would still remove this and check for `if bashio::fs.file_exists "${data_path}/config.yml" ; then`. Of course this also only works, if we implement the fail above if the file is there but not valid as suggested by me.

brenner-tobias: Also relevant on other locations

brenner-tobias (changes requested):
Great overall approach, I have added a couple of remarks / suggestions.

elcajon:
Thanks for the comments. I'm busy for the weekend, unfortunately. Will be able to look at it from Monday.

cloudflared/rootfs/etc/cont-init.d/cloudflared-config.sh
elcajon: The proposed logic will use the created cert and tunnel config from already implemented logic. 

1. Exactly: it will work if put there by user manually or created automatically with add-on logic
2. yep
3. In my opinion, the logic of custom config should only be used by advanced users anyway. Also, the config can only be created if the corresponding contents of the Cloudflared documentation are known. Thus, it should also be possible for the user to manually read the UUID from the tunnel.json and add it. Especially because all required files are located in the same directory. Don't you think we complicate the general add-on logic if we try to catch all cases? I wouldn't like to partially auto-complete a custom config with the add-on configuration.

PR #46, opened 2022-02-11
Add custom config.yml

elcajon:
# Proposed Changes

- > Add feature for using custom config.yml #43


brenner-tobias (changes requested):
I just did the review on the old PR, sorry for missing that you deleted it in between. Could you have a look there?

brenner-tobias (approved):

brenner-tobias:
Looks really good. Fair point regarding the tunnel UUID for the config, so let go with what you have implemented.

PR #48, opened 2022-02-14
Pass homeassistant log level to cloudflared

elcajon:
# Proposed Changes

- Pass existing config option `log_level` to cloudflared daemon

### DIFF
--- a/cloudflared/rootfs/etc/services.d/cloudflared/run
+++ b/cloudflared/rootfs/etc/services.d/cloudflared/run
@@ -17,7 +17,25 @@ else
     certificate="${data_path}/cert.pem"
     tunnel_name="$(bashio::config 'tunnel_name')"
 
+    # Map HomeAssistant log levels to Cloudflared
+    if bashio::config.exists 'log_level' ; then
+        case $(bashio::config 'log_level') in
+            "trace") cloudflared_log="debug";;
+            "debug") cloudflared_log="debug";;
+            "info") cloudflared_log="info";;
+            "notice") cloudflared_log="info";;
+            "warning") cloudflared_log="warn";;
+            "error") cloudflared_log="error";;
+            "fatal") cloudflared_log="fatal";;
+        esac
+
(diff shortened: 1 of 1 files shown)


cloudflared/rootfs/etc/services.d/cloudflared/run
```diff
@@ -23,7 +23,26 @@ else
...
+    else
+        cloudflared_log="info"
+    fi
+
     bashio::log.info "Connecting Cloudflared Tunnel..."
     bashio::log.debug "using ${config} config file"
-    exec cloudflared --origincert="${certificate}" --no-autoupdate tunnel --config="${config}" run "${tunnel_name}"
+    
+    exec cloudflared --origincert=${certificate} --no-autoupdate \
+        tunnel --config=${default_config} --loglevel="${cloudflared_log}" run "${tunnel_name}"
```
brenner-tobias, line 47: --loglevel property should also be included for the quick-tunnel in line 10, do you agree?
I also just realized that it is also missing the --no-autoupdate for the quick tunnel, so maybe you can fix that as well.

brenner-tobias (approved):
Perfect

brenner-tobias (approved):
+Protocol for quick_tunnel also makes a lot of sense

brenner-tobias:
after deploying to my "edge" HA instance, I am reconsidering if this is a good idea or if we should separate the Cloudflare log level from the HA one.
The reason is that the amount of Logs from Cloudflare e.g. on the debug level is practically making debugging for the add-on itself impossible. So maybe we should introduce another option `cloudflared_log-level` or similar?
What do you think @elcajon?

elcajon:
To be honest I have set up the logic to reduce the log entries for production use, to be able to focus on real errors related to the Cloudflare tunnel. 
I am not sure if we are making things unnecessarily complicated when adding another option just to differentiate log severity for development purposes. 😄 
What would you think if I just change the mapping so cloudflared will run with info log when ha runs with debug?

PR #50, opened 2022-02-15
Add initial documentation for strict ssl

### LINKED ISSUE #47: Export SSL certificates to /ssl
It would be nice if the Cloudflared add-on could export the SSL certificates to /ssl like the DuckDNS and Let's Encrypt add-on does.

For me, this would be useful for other server add-ons like Asterisk, which requires a SSL certificate regardless.

I don't know if it's possible, but if it is, it would be nice.

If it is possible, I believe that perhaps we could recommend people setup their http integration to enable HTTPS, so that traffic inside of the house also gets encrypted.

felipecrs:
# Proposed Changes

Add the following features:

##  Strict SSL mode

- [ ] Document 🏃
- [ ] Configuration validation
- [ ] Disable `noTLSVerify=true` if enabled

## Automatic generation of Origin CA certificate

- [ ] Document 🏃
- [ ] Configuration validation
- [ ] Implement through the [API](https://developers.cloudflare.com/ssl/origin-configuration/origin-ca#api-calls)

## Related Issues

Closes #47

### DIFF
--- a/cloudflared/DOCS.md
+++ b/cloudflared/DOCS.md
@@ -108,6 +108,10 @@ additional_hosts:
     service: "http://192.168.1.3:8080"
     disableChunkedEncoding: true
 nginx_proxy_manager: true
+strict_ssl: true
+generate_cert: true
+certfile: fullchain.pem
+keyfile: privkey.pem
 log_level: "debug"
 ```
 
@@ -143,6 +147,41 @@ additional_hosts:
 **Note**: _If you delete a hostname from the list, it will not be served
 anymore. Nevertheless, you should also manually delete the DNS entry from
 Cloudflare since it can not be deleted by the Add-on._
+
+### Option: `strict_ssl`
+
+When enabled, the tunnel in Cloudflare will be set with the [_Full (strict)_ SSL/TLS encryption mode](https://developers.cloudflare.com/ssl/origin-configuration/ssl-modes#full-strict) to provide a higher security.
+
+In this mode, the local connection needs to be stablished over HTTPS, presenting a certificate issued by a publicly trusted certificate authority (like using the [Let's Encrypt addon](https://github.com/home-assistant/addons/tree/master/letsencrypt#home-assistant-add-on-letsencrypt)), or a [Cloudflare Origin CA](https://blog.cloudflare.com/cloudflare-ca-encryption-origin/) one (which [the addon can generate for you automatically](#option-generatecert)).
+
+Here is an example on how to configure the Home Assistant HTTP server to use HTTPS with the appropriate certificate:
+
+```diff-json
+// /config/configuration.yaml
+
+http:
+  use_x_forwarded_for: true
+  trusted_proxies:
+    - 172.30.33.0/24
++ ssl_certificate: /ssl/fullchain.pem
++ ssl_key: /ssl/privkey.pem
+```
+
+### Option: `generate_cert`
+
+When enabled, the addon will generate, retrieve and export a Cloudflare Origin CA issued certificate for you in the `/ssl` folder.
+
+> Please note that the CloudFlare Origin CA certificates are only valid for using between the Cloudflare Tunnel and the local server, as it is not issued by a publicly trusted certificate authority.
+>
+> If for some reason, you need publicly trusted certificates, you should disable this option and something like the [Let's Encrypt addon]([Let's Encrypt addon](https://github.com/home-assistant/addons/tree/master/letsencrypt#home-assistant-add-on-letsencrypt)) instead.
+
+#### Option: `certfile`
+
+The path where the generated public certificate will be exported to, relative to `/ssl`.
+
+#### Option: `keyfile`
+
+The path where the generated certificate private key will be exported to, relative to `/ssl`.
 
 ### Option: `catch_all_service`
 
--- a/cloudflared/config.yaml
+++ b/cloudflared/config.yaml
@@ -21,10 +21,20 @@ map:
 options:
   external_hostname: null
   tunnel_name: "homeassistant"
+  strict_ssl: false
+  generate_cert:
+    enabled: false
+    certfile: fullchain.pem
+    keyfile: privkey.pem
   additional_hosts: []
 schema:
   external_hostname: str
   tunnel_name: str
+  strict_ssl: bool
+  generate_cert:
+    enabled: bool
+    certfile: str
+    keyfile: str
   additional_hosts:
     - hostname: str
       service: str
--- a/cloudflared/rootfs/etc/cont-init.d/10-cloudflared-config.sh
+++ b/cloudflared/rootfs/etc/cont-init.d/10-cloudflared-config.sh
@@ -113,7 +113,7 @@ createCertificate() {
     bashio::log.notice
     cloudflared tunnel login
 
-    bashio::log.green "Authentication successfull, moving auth file to config folder"
+    bashio::log.green "Authentication successfull, moving auth file to the '${data_path}' folder"
 
     mv /root/.cloudflared/cert.pem "${data_path}/cert.pem" || bashio::exit.nok "Failed to move auth file"
 


felipecrs:
@brenner-tobias, @elcajon I wrote the documentation draft. Are we aligned so far?

cloudflared/DOCS.md
```diff
@@ -144,6 +148,41 @@ additional_hosts:
 anymore. Nevertheless, you should also manually delete the DNS entry from
 Cloudflare since it can not be deleted by the Add-on._
 
+### Option: `strict_ssl`
```
brenner-tobias, line 151: I don't think this option is needed and the description is indeed wrong:

1. I think once we decide to generate_cert, we should automatically remove the option `no-tls-verify: true`. So no need for a separate option here.
2. As I mentioned before, SSL strict can already be used in Cloudflare. The option `no-tls-verify: true` does not affect this. Strict only means, that the connection to the origin server are encrypted. Since we are using the encrypted tunnel, it seems like Cloudflare is fine with this.

cloudflared/DOCS.md
```diff
@@ -108,6 +108,10 @@ additional_hosts:
     service: "http://192.168.1.3:8080"
     disableChunkedEncoding: true
 nginx_proxy_manager: true
+strict_ssl: true
+generate_cert: true
+certfile: fullchain.pem
```
brenner-tobias, line 113: Dont't think we need to make the name editable. we can simply both of them name "fullchain_cloudflare.pem" and "privkey_cloudflare.pem" or something like that, no no overrides are to be expected

cloudflared/DOCS.md
```diff
@@ -144,6 +148,41 @@ additional_hosts:
...
+Here is an example on how to configure the Home Assistant HTTP server to use HTTPS with the appropriate certificate:
+
+```diff-json
+// /config/configuration.yaml
+
+http:
+  use_x_forwarded_for: true
+  trusted_proxies:
+    - 172.30.33.0/24
++ ssl_certificate: /ssl/fullchain.pem
```
brenner-tobias, line 166: See comment of certfile names at the beginning. I would add here the respecting name with "...cloudflare.pem" or whatever and go with that

cloudflared/DOCS.md
```diff
@@ -144,6 +148,41 @@ additional_hosts:
...
++ ssl_key: /ssl/privkey.pem
+```
+
+### Option: `generate_cert`
+
+When enabled, the addon will generate, retrieve and export a Cloudflare Origin CA issued certificate for you in the `/ssl` folder.
+
+> Please note that the CloudFlare Origin CA certificates are only valid for using between the Cloudflare Tunnel and the local server, as it is not issued by a publicly trusted certificate authority.
+>
+> If for some reason, you need publicly trusted certificates, you should disable this option and something like the [Let's Encrypt addon]([Let's Encrypt addon](https://github.com/home-assistant/addons/tree/master/letsencrypt#home-assistant-add-on-letsencrypt)) instead.
```
brenner-tobias, line 176: I would not mention that here or anywhere. This only confuses the average user. We are with our Add-on using very secure e2e encryption already, so no need to disturb the average user here. Pros with the respecting need for other certificates will understand with the previous paragraph.

cloudflared/config.yaml
```diff
@@ -21,10 +21,20 @@ map:
 options:
   external_hostname: null
   tunnel_name: "homeassistant"
+  strict_ssl: false
```
brenner-tobias, line 24: As mentioned above, I would reduce all these options to just a normal optional option: 

```
schema:
   ...
   generate_cert:bool?
   ...
```

elcajon, line 24: I would suggest to add this option for homeassistant (`external_hostname`) and for `additional_hosts[]` to separate the different services. This would require to restructure the possible options and schema. e.g.:

```yaml
...
options:
  homeassistant:
    external_hostname: ha.example.com
    strict_ssl: false
  additional_hosts:
    - hostname: example.com
       service: example
       disableChunkedEncoding: false
       strict_ssl: false
...
```

What do you think?

cloudflared/rootfs/etc/cont-init.d/10-cloudflared-config.sh
```diff
@@ -113,7 +113,7 @@ createCertificate() {
     bashio::log.notice
     cloudflared tunnel login
 
-    bashio::log.green "Authentication successfull, moving auth file to config folder"
+    bashio::log.green "Authentication successfull, moving auth file to the '${data_path}' folder"
```
brenner-tobias, line 116: Like the suggestion, would slightly change again:

`bashio::log.green "Authentication successful, moving auth file to the custom data folder: '${data_path}'"`

cloudflared/config.yaml
```diff
@@ -21,10 +21,20 @@ map:
...
+  generate_cert:
+    enabled: false
+    certfile: fullchain.pem
+    keyfile: privkey.pem
   additional_hosts: []
 schema:
   external_hostname: str
   tunnel_name: str
+  strict_ssl: bool
```
elcajon, line 33: I think we should make all those parameters optional by adding the `schema` part only (like `disableChunkedEncoding`)

brenner-tobias (changes requested):
Thanks a lot, the general approach looks good, though I think we should simplify the options to offer. See detailed comments.

cloudflared/DOCS.md
```diff
@@ -144,6 +148,41 @@ additional_hosts:
...
+
+#### Option: `certfile`
+
+The path where the generated public certificate will be exported to, relative to `/ssl`.
+
+#### Option: `keyfile`
+
+The path where the generated certificate private key will be exported to, relative to `/ssl`.
```
elcajon, line 184: Are we accepting paths as `keyfile` and `certfile` value or only the file names?
I guess we need this checked within the `schema` part.

cloudflared/DOCS.md
```diff
@@ -144,6 +148,41 @@ additional_hosts:
...
+  use_x_forwarded_for: true
+  trusted_proxies:
+    - 172.30.33.0/24
++ ssl_certificate: /ssl/fullchain.pem
++ ssl_key: /ssl/privkey.pem
+```
+
+### Option: `generate_cert`
+
+When enabled, the addon will generate, retrieve and export a Cloudflare Origin CA issued certificate for you in the `/ssl` folder.
```
elcajon, line 172: Should we generate a new origin certificate or just extract the origin zone certificate from the already generated cert.pem file (after cloudflared login)? 
I would suggest to extract the cert and key from cert.pem, otherwise we would need to pas another Cloudflare API token (afaik) to the add-on.

cloudflared/rootfs/etc/cont-init.d/10-cloudflared-config.sh
elcajon: This variable doesn't necessarily contain the custom data folder. If no data_folder option is provided it will simply contain /data. So I think the log is correct. But I would change method to bashio::log.**info**

cloudflared/DOCS.md
elcajon: I agree in general. One further question to 1.)

I am not sure if we really can remove noTLSVerify: true after setting up Cloudflare zone certificate for e.g. homeassistant. We call homeassistant server with https://homeassistant:8123 and assign our origin certificate issued for *.example.com and example.com. I assume the TLS check will fail as hostname and certificate host will mismatch. 
I think to avoid this error we have to add another ingress rule [originServerName](https://developers.cloudflare.com/cloudflare-one/connections/connect-apps/configuration/configuration-file/ingress#originservername), and pass the Cloudflare zone name (e.g. example.com). To be honest, I have not looked at this more closely.

cloudflared/DOCS.md
brenner-tobias: I guess we have to get an new certificate from Cloudflare, with a long enough validity for this to actually make sense. I think we can use the certificate for the communication with the Cloudflare API though. Did not really look into it to be honest, so if it is working with extracting everything from what we have already, happy to do so

cloudflared/config.yaml
brenner-tobias: agree for the additional_hosts part. For the homeassisatant part the question remains if we need it or not which depends on whether we want to link that to the creation of the certificate as suggested by me or not

cloudflared/DOCS.md
brenner-tobias: You are completely right, very important point. My understanding is also be that we need to add the origin server name as another option of `originRequest` here and the check should work only after that.

brenner-tobias: One thing to clarify here @elcajon: do you agree that we should imply strict ssl for the HA connection once the user sets `generate_cert ` and only have the option for the `additional_hosts`?

elcajon: I would agree on that. We should then also add the originRequest - originServerName automatically analogous to hostname_external for HA, right?

cloudflared/config.yaml
elcajon: In relation to the discussion in the other comment, the scheme would then look like this in my opinion
```yaml
...
options:
  external_hostname: ha.example.com
  additional_hosts:
    - hostname: example.com
      service: example
      disableChunkedEncoding: false
      noTLSVerify: false (set default per program logic) (schema: bool?)
      originServerName: example.com (schema: str?) 
...
```

brenner-tobias: Two things regarding this:
1. disableChunkedEncoding would also be optional (as it is at the moment)
2. I do not think that we need originServerName. This should always be the same as the hostname, right? Unless we also want to enable the user to work with even other certificates then the Cloudflare ones....maybe leave that for the time being and add it at a second stage?

So to summarise, the definition in the config would be like this:


```
options:
  external_hostname: null
  tunnel_name: "homeassistant"
  additional_hosts: []
schema:
  external_hostname: str
  tunnel_name: str
  additional_hosts:
    - hostname: str
      service: str
      disableChunkedEncoding: bool?
      noTLSVerify: bool?
  catch_all_service: str?
  nginx_proxy_manager: bool?
  quick_tunnel: bool?
  log_level: list(trace|debug|info|notice|warning|error|fatal)?
  reset_cloudflared_files: bool?
  data_folder: list(config|share|ssl)?
  custom_config: bool?
  generate_cert: bool?
```

elcajon: I agree, good point.

cloudflared/DOCS.md
felipecrs: `noTLSVerify: true` does not make sense to fail if the hostname mismatches the certificate. Cloudflared is meant to connect origin servers (obviously not through the domain name, but by the hostname/IP in network). So, just wondering, have someone tested this? I believe the verification that `noTLSVerify: true` is to verify whether the origin certificate matches the domain of the final domain name, not the origin domain name.

It also checks if the CA that issued the certificate is either a public valid one like Let's Encrypt or Cloudflare Origin's.

But besides that, I do agree 100% that the option should not be named `strict_ssl`. I just checked and even though the addon is talking to my Home Assistant over HTTP (not HTTPS), I could enable Strict SSL in the Cloudflare panel, and it works.

So, bottom line is that Strict SSL has absolutely nothing to do with what I proposed.

Perhaps I should change it to something like `tls_verify`, which would:

1. Be false by default
2. Be always true if "generate_cert" is enabled
3. But can also be enabled manually, in case user is using certificates generated by other means, like Let's Encrypt.

So, by this I think that the option should be kept as opposed to be removed as initially proposed by @brenner-tobias.

WDYT?

felipecrs: Oh, and also, for now everything in this PR is only valid for HA. We can think about additional hostnames later. As suggested by @brenner-tobias;

cloudflared/DOCS.md
felipecrs: Hm... I disagree.

Every HTTPS guide in Home Assistant suggests using `/ssl/fullchain.pem` and `/ssl/privkey.pem`.

That's where the Let's Encrypt addon exports certificates to, and the DuckDNS addon as well.

Also, many addons comes configured by default to use the certificates placed in these locations as well. Changing them to something else would need users to manually tweak the (good) defaults of the existing addons, like Nginx proxy, TasmoAdmin, or any other HASS Community addon which serves HTTPS.

Also, it's unlikely for people to have two different addons generating certificates for them, like having both Let's Encrypt and Cloudflared.

But, if the user desires so for whatever reason, this is why allowing to change the default value would be good.

cloudflared/DOCS.md
felipecrs: Agreed.

cloudflared/rootfs/etc/cont-init.d/10-cloudflared-config.sh
felipecrs: Well, this is totally off-topic. I'll remove this change from this PR and put it in another.

felipecrs: https://github.com/brenner-tobias/addon-cloudflared/pull/51

cloudflared/config.yaml
felipecrs: How about we leave `additional_hosts` for a second PR?

cloudflared/config.yaml
felipecrs: Well. The benefit is:

- User does not have even have think about it in the configuration page

The caveat is:

- User will need to open the DOCS to check what is the default value, which otherwise would be on front the configuration itself

So, as you both seem to agree that it should be transparent/optional, I will do it.

cloudflared/DOCS.md
felipecrs: I did not dig into it already, so I don't know how the authentication with the Cloudflare API will be made yet. I will share more details once I start working on it.

cloudflared/DOCS.md
elcajon: I did a quick check with my ProxmoxPVE host. After adding private key and certificate (both extracted from cert.pem) to the host and setting noTLSVerify:false the connection fails with Cloudflare SSL full (strict). 
After adding the originServerName, connection could be established. 

This is the working configuration part: 

```yaml
- hostname: proxmox.example.com
  service: https://10.10.0.2:8006
  originRequest:
    disableChunkedEncoding: true #required for Proxmox, not needed for SSL in general
    noTLSVerify: false
    originServerName: example.com
```

brenner-tobias: So this means, that once you add the certificates, they have to be valid for the given service or you have to disable TLS verify or tell Cloudflare which originServerName to expect, right?
Thinking about the strict ssl flag, I am torn back and forth. Now talking about the HA service: I think there is no case where anyone would create a certificate another way then using the one from Cloudflare. In any case, this would not be best practice, sind we are operating in the Cloudflare Ecosystem and also the automatic renewal would not work so easily. Therefor, I still think that we should try to keep it simple and do not offer that option, unless you see any other cases I am missing.

cloudflared/DOCS.md
brenner-tobias: Mhm fair enough, I am still not sure about the config option. Maybe another option would be so set the certificates to the standard-names as suggested by you (the point you make about the default config of other add-ons does make a lot of sense), but do not override and fail if there are already files with that name in the folder?

cloudflared/config.yaml
brenner-tobias: I think we should set generate certificate to 1 anyways as a default, so that the average user will just go with it and have the SSL support as well. Therefore, it would be a pretty uncommon edge case, if you do not want to create the certificate.
If you already have certificate files in SSL, it will print an error on start-up and point you to the respecting solution / configuration option to disable for whatever reason.

What do you think?

elcajon: > I think we should set generate certificate to 1 anyways as a default, so that the average user will just go with it and have the SSL support as well.

Would that mean generating the certificates as default behavior and setting the ingress settings to noTLSVerify=false and originServerName=external_hostname? This would imply that every user who uses the add-on also configures all services with the certificate by default?

cloudflared/DOCS.md
felipecrs: > do not override and fail if there are already files with that name in the folder?

Perhaps the whole `generate_cert` should be disabled by default, and if users intentionally enable it, they will read in the documentation that it will overwrite existing files if they exist already.

By the way, overwriting already happens in the Let's Encrypt addon for a good reason: it automatically renews the certificate if the expiring date is approaching.

felipecrs: Since HTTPS is not actually needed for enabling Strict SSL in Cloudflare (as you know, I initially thought the opposite), I think it's reasonable that we should not bother people about asking them to enable HTTPS in their HA, unless they want to. Thus, `generate_cert` would be disabled by default.

cloudflared/DOCS.md
felipecrs: I think we don't need to bother about automatic renewal for now... the certificates issued by Cloudflare Origin CA are by default valid for 15 years. It can be a later improvement.

felipecrs: And, generally speaking, I find that giving users more control over the underlying tools is always better, so they can extend their use cases if they want.

But I do agree that we should not bother offering every single Cloudflared configuration key. As we already allow users to set their own custom cloudflared configuration through the _Option: custom_config_, I think it may actually be a better fit to let users specify there if they want.

The only problem, however, is that the addon will always override `noTlsVerify`, thus not allowing the users to set it effectively. Am I right?

brenner-tobias: It will not override anything in the config if the users uses a custom config. So we can completely disregard that part and only focus on the "average" user here.

cloudflared/DOCS.md
brenner-tobias: Mhm not sure about this one. So they have to change something in the config anyways (so HA accepts the packages from Cloudflare). That means, that they just have to copy more into their config, which is not more work. At the same time, we would make it the standard case, that they have SSL also in their local network, which I feel is something that quite some people don't like about the current standard approach in this Add-on. So why not just make it standard to have SSL?

cloudflared/config.yaml
brenner-tobias: Yes and no. So setting everything you described yes, but only for the Ingress rule of HomeAssistant. For the additional hosts, we would have the separate flag to decide.

brenner-tobias:
@felipecrs are you still working on this? I think it makes sense to update with the current main since we introduced a new functionality which might lead to merging conflicts in the future.

felipecrs:
Hey @brenner-tobias, thanks for the follow-up.

Not really since our last iteration of comments. In fact, I seriously doubt about the value that this would bring, considering that you don't need this to turn on Strict SSL in Cloudflare.

Also, I personally don't see a use case to enable HTTPS locally any longer. Even so, any kind of certificate, publicly valid, issued by Cloudflare Origin CA, or even self-signed would work.

The only benefit is that, if it's a Cloudflare Origin CA, we could disable `noTLSVerify`.

Still, considering my reasonings above, if it was up to me alone, I don't think it worth it. If neither of us three consider using it, perhaps we can close it and reconsider if someone else who really needs ask us.

What do you guys think?

PR #64, opened 2022-03-06
Bugfix reading tunnel name from tunnel.json

elcajon:
# Proposed Changes

Upstream logic changed when running `cloudflared tunnel create`.
`json.conf` no longer contains the `TunnelName` key.

Adapting add-on logic to gather tunnel name using Cloudflare API and tunnel id. 

## Related Issues

### DIFF
--- a/cloudflared/rootfs/etc/cont-init.d/10-cloudflared-config.sh
+++ b/cloudflared/rootfs/etc/cont-init.d/10-cloudflared-config.sh
@@ -141,7 +141,8 @@ hasTunnel() {
     # Check if tunnel name in file matches config value
     bashio::log.info "Checking if existing tunnel matches name given in config"
     local tunnel_name_from_file
-    tunnel_name_from_file="$(bashio::jq "${data_path}/tunnel.json" .TunnelName)"
+    tunnel_name_from_file=$(cloudflared --origincert="${data_path}/cert.pem" tunnel \
+        list --output="json" --id="${tunnel_uuid}" | jq -er '.[].name')
     bashio::log.debug "Tunnnel name read from file: $tunnel_name_from_file"
     if [[ $tunnel_name != "$tunnel_name_from_file" ]]; then
         bashio::log.warning "Tunnel name in file does not match config, removing tunnel file"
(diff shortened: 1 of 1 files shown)


brenner-tobias (changes requested):

cloudflared/rootfs/etc/cont-init.d/10-cloudflared-config.sh
```diff
@@ -138,12 +138,13 @@ hasTunnel() {
...
-    # Check if tunnel name in file matches config value
+    # Get tunnel name from Cloudflare API by tunnel id and chek if it matches config value
     bashio::log.info "Checking if existing tunnel matches name given in config"
-    local tunnel_name_from_file
-    tunnel_name_from_file="$(bashio::jq "${data_path}/tunnel.json" .TunnelName)"
-    bashio::log.debug "Tunnnel name read from file: $tunnel_name_from_file"
-    if [[ $tunnel_name != "$tunnel_name_from_file" ]]; then
+    local existing_tunnel_name
+    existing_tunnel_name=$(cloudflared --origincert="${data_path}/cert.pem" tunnel \
+        list --output="json" --id="${tunnel_uuid}" | jq -er '.[].name')
```
felipecrs, line 145: I think you can use `bashio::jq` here, no?

felipecrs:
```suggestion
        list --output="json" --id="${tunnel_uuid}" | bashio::jq -e '/dev/stdin' '.[].name')
```

Not sure, but it may work.

felipecrs: Nevermind. I think it does not accept the -e.

felipecrs (dismissed):

cloudflared/rootfs/etc/cont-init.d/10-cloudflared-config.sh
elcajon: I also tried around with bashio::jq, I guess this would work, but not sure.
```
bashio::jq "$(cloudflared --origincert="${data_path}/cert.pem" tunnel \
    list --output="json" --id="${tunnel_uuid}")" '.[].name'
```
I think if we want to add more features in the future, we'll have to restructure all the code either way. Things getting confusing when working with only one file

brenner-tobias (approved):

PR #65, opened 2022-03-07
Add Cloudflare Warp routing

### LINKED ISSUE #62: Support Local Network Routing Through Warp Client
Thanks for creating a great add on! 

**Is your feature request related to a problem? Please describe.**
I’m unable to use the Home Assistant apps through the tunnel when authentication is turned on. I’m able to login through the Cloudflare one time password wall, but the apps doesn’t seem to accept the resulting cookie.

**Describe the solution you'd like**
A workaround would be to support the Cloudflare Zero Trust companion app, that effectively creates a VPN to your local network from your device. I have this up an running on a stand-alone device in my LAN, and it works great. It would involve adding this lines to the Cloudflare config

```
warp-routing:
  enabled: true
```

Adding a route to your tunnel like this: 

`cloudflared tunnel route ip add 192.168.XX.0/24 tunnelname`

And finally remove your LAN IP subnet in Cloudflare’s Split Tunnels settings.

See https://digaround.cloud/homeassistant-secure-access-with-cloudflare-warp/ for full context.

elcajon:
# Proposed Changes

I consider this as WIP as there is definitely room for improvements. Feel free to criticize. 😉 

```
warp_enable: true
warp_routing:
  - 10.0.0.0/24
  - 10.10.0.0/24
warp_reset: true
```

This PR introduces three new (optional) configuration options. 

`warp_enable: true` Will add the necessary options to json.config to allow warp functionality. Furthermore if `warp_routing`isn't specified, the existing network interfaces will be read and routing will be set up for cloudflare tunnel.

If `warp_routing: ...`is set, the given hosts will be used for routing and host interfaces will be ignored. 

`warp_reset: true` will remove all warp related add-on configuration options and delete all tunnel related routes. 

Open points: 

- Config options check (e.g. warp_routing is set and warp_enable isn't...)
- How and which routes to delete? For now I delete all routes assigned to the tunnel before setting the configured ones up. Is there a more common way?
- ...

## Related Issues

> closes #62


cloudflared/rootfs/etc/cont-init.d/10-cloudflared-config.sh
```diff
@@ -352,6 +360,112 @@ hasCustomConfig() {
...
+# ------------------------------------------------------------------------------
+createRoute() {
+    if ! bashio::config.has_value 'warp_enable' || bashio::config.false 'warp_enable' ; then
+        return "${__BASHIO_EXIT_OK}"
+    fi
+
+    # Remove already linked routes 
+    bashio::log.info "Removing already configured routes for tunnel ${tunnel_name}"
+    # Get routes linked to tunnel id
+    existing_tunnel_routes=$(cloudflared --origincert="${data_path}/cert.pem" \
```
brenner-tobias, line 374: I think we can take this deletion part and extract it to it's own method and then share it with "warp_reset". Does that make sense?

cloudflared/rootfs/etc/cont-init.d/10-cloudflared-config.sh
```diff
@@ -352,6 +360,112 @@ hasCustomConfig() {
...
+    done
+
+}
+
+# ------------------------------------------------------------------------------
+# Reset routes and config options
+# ------------------------------------------------------------------------------
+warp_reset() {
+    # Remove already linked routes 
+    bashio::log.info "Removing already configured routes for tunnel ${tunnel_name}"
```
brenner-tobias, line 449: see comment above, can be extracted to own method from my pov

cloudflared/rootfs/etc/cont-init.d/10-cloudflared-config.sh
```diff
@@ -400,10 +514,17 @@ main() {
...
+
+    if bashio::config.true 'warp_reset' ; then
+        warp_reset
+    fi
+
     createConfig
 
     createDNS
 
+    createRoute
```
brenner-tobias, line 526: I think we should stay in the same pattern here and check the config before calling the method and not check in the method "createRoute" itself and return if not set. Obviously works both, it should just be the same for all the options.

cloudflared/rootfs/etc/cont-init.d/10-cloudflared-config.sh
```diff
@@ -352,6 +360,112 @@ hasCustomConfig() {
     bashio::exit.nok "No custom config found: ${data_path}/config.yml please create custom config file or remove 'custom_config' option"
 }
 
+# ------------------------------------------------------------------------------
+# Create route for local IPs if Warp is enabled
+# ------------------------------------------------------------------------------
+createRoute() {
+    if ! bashio::config.has_value 'warp_enable' || bashio::config.false 'warp_enable' ; then
```
brenner-tobias, line 367: See comment on bottom. I think we should move that check to the main method before calling it

cloudflared/config.yaml
```diff
@@ -36,3 +36,7 @@ schema:
   reset_cloudflared_files: bool?
   data_folder: list(config|share|ssl)?
   custom_config: bool?
+  warp_enable: bool?
+  warp_routes:
```
brenner-tobias, line 40: I cannot test right now, but from my understanding, once you add an option with an array like this, the array always needs to be present in the config then, as an empty array. That's why we have to add "additional_hosts: []" always, even if we don't need it.
Did you check, if the add-on continues to work without that option? If not as I expect, I think we should add the empty array above in the options part to not break the standard flow.

brenner-tobias (commented):
Great work. Added some initial thoughts

elcajon:
Good suggestions, I will implement the changes.

cloudflared/config.yaml
elcajon: I have tested this and it works, I would guess because it is just a string list and not a list nesting objects. 

See schema from [Dev Docs](https://developers.home-assistant.io/docs/add-ons/configuration)

```
message: str
logins:
  - username: str
    password: str
random:
  - "match(^\\w*$)"      <--- 
link: url
size: "int(5, 20)"
count: float
not_need: "str?"
```

brenner-tobias: ok, perfect

cloudflared/rootfs/etc/cont-init.d/10-cloudflared-config.sh
```diff
@@ -352,6 +360,106 @@ hasCustomConfig() {
     bashio::exit.nok "No custom config found: ${data_path}/config.yml please create custom config file or remove 'custom_config' option"
 }
 
+# ------------------------------------------------------------------------------
+# Create route for local IPs if Warp is enabled
```
brenner-tobias, line 364: Should be fixed with actual comment of deleteRoutes()

brenner-tobias (approved):

brenner-tobias (approved):

PR #80, opened 2022-03-23
Add check for valid hostname

brenner-tobias:
# Proposed Changes

> Add a Regex check for the external_hostname and hostname of additional_hosts, so that users get a proper error message if they e.g. include https or the port in the hostname


elcajon (approved):
👍

PR #95, opened 2022-05-09
Add remote managed tunnel functionality

elcajon:
# Proposed Changes

This PR introduces the `tunnel_token` config option. When set, all other options will be ignored and the dashboard managed tunnel will be connected. 

# ToDo's

- [X] I guess there is still room for improvement when it comes to the documentation. 
- [x]  Maybe we should also mark this as breaking change (even if all existing configuration should run without interruption) as it may enables more people to use this add-on thanks to a much easier configuration process. Also changes made via the dashboard are available instant without the need to restart the add-on.


brenner-tobias:
You want to go with "Breaking change" in order to release it as a new major version, correct? If so, we can also use the label "major", which would be my preferred solution.

cloudflared/rootfs/etc/services.d/cloudflared/run
```diff
@@ -15,8 +15,13 @@ if bashio::config.true 'quick_tunnel'; then
     bashio::log.info "Connecting Cloudflared Quick Tunnel..."
     exec cloudflared --no-autoupdate \
     tunnel --loglevel="${CLOUDFLARED_LOG}" --url "${ha_service_protocol}://homeassistant:$(bashio::core.port)"
-else
 
+elif bashio::config.has_value 'tunnel_token' ; then
+    tunnel_token=$(bashio::config 'tunnel_token')
+    exec cloudflared --no-autoupdate \
```
brenner-tobias, line 21: I assume that the tunnel option also means ignoring any config files that might exist. Cannot verify at the moment though, but let's go with it

brenner-tobias (approved):
Awesome

brenner-tobias:
@elcajon feel free to merge if the label major is fine for you. Also, I suggest we wait with releasing this for a couple of days to test it properly

elcajon:
Everything seems to work fine so far, I guess we are good to go. I would say we should release the new version asap, as there is also the latest cloudflared version in it. What do you think @brenner-tobias?

brenner-tobias:
Agreed, looks good on my end too. Feel free to release 👍🏻

Issue #99, opened 2022-05-21
Tunnel with name already exists


iosoft:
**Describe the bug**
I use a Cable Internet connection. Sometimes connection gets cut for few minutes or seconds.
I have noticed that, in some _rear cases_, this plugin stops working totally. Can't create the tunnel any more.
Device restart or Add-on restart doesn't work!
**I need to Delete the Tunnel from dash.teams.cloudflare.com and then restart the Add-on/Full Device to restart the tunnel.**

**Log Error Message**
> FATAL: Failed to create tunnel. Please check the Cloudflare Teams Dashboard for an existing tunnel with the name homeassistant and delete it: https://dash.teams.cloudflare.com/ Access / Tunnels.

**Expected behavior**
It should re-establish the broken connection/tunnel automatically.
Pls fix this annoying bug for this 99.99999% perfect Add-on.

**Add-On / Home Assistant Information (please complete the following information):**
- Add-On Version (2.0.0) latest
- Supervisor Version (2022.05.2)
- Core Version (2022.5.5)
- Operating System (e.g. Home Assistant OS 8.1)
- Devices (e.g. Raspberry Pi 3 Model B)

**Log**

> [cont-init.d] 10-cloudflared-config.sh: executing... 
> failed to create tunnel: Create Tunnel API call failed: tunnel with name already exists
> [14:57:49] FATAL: Failed to create tunnel.
>     Please check the Cloudflare Teams Dashboard for an existing tunnel with the name homeassistant and delete it:
>     https://dash.teams.cloudflare.com/ Access / Tunnels
> [cont-init.d] 10-cloudflared-config.sh: exited 1.
> [cont-finish.d] executing container finish scripts...
> [cont-finish.d] 99-message.sh: executing... 
> -----------------------------------------------------------
>                 Oops! Something went wrong.
> 
>  We are so sorry, but something went terribly wrong when
>  starting or running this add-on.
>  
>  Be sure to check the log above, line by line, for hints.
> -----------------------------------------------------------
> [cont-finish.d] 99-message.sh: exited 0.

dutch-erik:
Same here:
`Error 1033  Ray ID: 70efb6c62d39422a • 2022-05-21 19:28:13 UTC
Argo Tunnel error `

`[21:28:09] INFO: Creating new tunnel...
failed to create tunnel: Create Tunnel API call failed: tunnel with name already exists
[21:28:15] FATAL: Failed to create tunnel.
[...]

PR #132, opened 2022-07-29
Log error message if tunnel name doesn't fit 

elcajon:
# Proposed Changes

#99 - Tunnel with name already exists

The error is occurring more frequently according to the HA community forum.
I can still imagine that the API call below (line 162) is leading to the problem and the current coding is overriding the actual problem by deleting the tunnel file. I would like to bypass the automated deletion of the file via an error message to make the error (if I'm right) visible. 

https://github.com/brenner-tobias/addon-cloudflared/blob/ae091e84705e34e5647b5258648eec9e278ff3cf/cloudflared/rootfs/etc/s6-overlay/scripts/10-cloudflared-config.sh#L160-L169

## Related Issues


elcajon:
needs #133 for CI

brenner-tobias (approved):

Issue #125, opened 2022-07-18
Cloudflare fail to reconnect after internet lost


krzysiek2788:
### The problem

Hi,

I findout that addon has issue with reconnecting back to Cloudflare tunnel if server lost connection. It looks like Add-on is unable to reset tunnel connection automatically, after I restart add-on all is working fine.

### What version of Cloudflared has the issue?

2.0.4

### What type of installation are you running?

Home Assistant OS

### Add-on YAML Configuration

```yaml
additional_hosts: []
external_hostname: xxxxxxxxxx
tunnel_name: xxxxxxxxx
tunnel_token: ''
data_folder: ssl
```

### Anything in the logs that might be useful for us?

```txt
022-07-16T13:12:39Z ERR Unable to establish connection with Cloudflare edge error="TLS handshake with edge error: read tcp 172.30.33.5:48064->198.41.192.47:7844: read: connection reset by peer" connIndex=3 ip=198.41.192.47
2022-07-16T13:12:39Z ERR Serve tunnel error error="TLS handshake with edge error: read tcp 172.30.33.5:48064->198.41.192.47:7844: read: connection reset by peer" connIndex=3 ip=198.41.192.47
2022-07-16T13:12:39Z INF Retrying connection in up to 1s seconds connIndex=3 ip=198.41.192.47
2022-07-16T13:12:39Z ERR Unable to establish connection with Cloudflare edge error="TLS handshake with edge error: read tcp 172.30.33.5:53778->198.41.192.67:7844: read: connection reset by peer" connIndex=1 ip=198.41.192.67
```

### Steps to reproduce the issue

Disconnect internet for some time and Cloudflare add-on couldn't estabilished connection

elcajon:
Thanks for pointing this out. I've experienced the same problem a couple of times during the last weeks (due to ISP outage and router firmware updates...). I have found an (old and) open upstream [issue](https://github.com/cloudflare/cloudflared/issues/306) for this problem.

For now the only quick workaround I can figure out is to increase the `--retries` value to e.g. 20.

A lasting option would be to implement a monitoring service inside the add-on which triggers a restart if the daemon stucks. I guess we could grab the values from the included metrics server.

krzysiek2788:
Maybe you can implement action according to logs it could look like this:
1. If connection failed -> ping google.com
2. ping untill it success then restart add-on

This is just a simple idea :)

Beeez:
Seeing the same issue here

[...]

PR #139, opened 2022-08-03
Healthcheck for HA cloudflare tunnel if connection is dropped

elcajon:
# Proposed Changes

As discussed in #125 this PR implements a small healthcheck using the built-in metrics server.

To summarize the function: 
- Waiting for metrics server to be available (check if cloudflared is up and running, independent of connection status to clouflare edge servers)
- Check if status is 200 (= connected), all fine
- Otherwise increment counter
- Counter = 24 --> reinitialize cloudflared service (send restart command to s6-overlay)
- wait 5 sec

The threshold with 5 sec sleep and 24 loops is set to 120 seconds. I guess this should fit to most environments. 

I was reluctant to include this as a configuration option because the use cases are limited and the add-on configuration is already relatively complex  
What we can think about would be a general option to enable/disable all healthcheck functionality.

## Related Issues

> #125

### DIFF
--- a/cloudflared/rootfs/build.sh
+++ b/cloudflared/rootfs/build.sh
@@ -40,4 +40,4 @@ chmod +x /usr/bin/cloudflared
 
 # Add needed packages
 apk add --no-cache \
-    yq=4.25.1-r2
+    yq=4.25.1-r3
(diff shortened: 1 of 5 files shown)


brenner-tobias (approved):
This is awesome. Let's test it for a wile in the Edge repo. Thanks a lot!

brenner-tobias:
since we now have the metrics server, what do you think about also adding it as ingress so you can watch the metrics server from HA @elcajon ?

elcajon:
I also thought about that recently, do you know a suitable template or framework to present the data or do we have to start from scratch?

brenner-tobias:
not really, I was just considering making the standard html of the metrics server available via an ingress rule. Though this is just text and does not look good at all, do you have a better idea?

elcajon:
Not by heart, I will check some resources and Google.

Issue #141, opened 2022-08-11
Add-on stopped after couple of service restart


krzysiek2788:
### The problem

After latest updates of add-on which resolved issue with not reconnecting with cloudflare after internet loose, know it stopping whole add-on after couple of service restart

### What version of Cloudflared has the issue?

2.0.7

### What type of installation are you running?

Home Assistant OS

### Add-on YAML Configuration

```yaml
additional_hosts: []
external_hostname: xxxx
tunnel_name: xxxx
tunnel_token: ''
data_folder: ssl
```

### Anything in the logs that might be useful for us?

```txt
2022-08-11T03:53:30Z INF Switching to fallback protocol http2 connIndex=1 ip=198.41.192.37
2022-08-11T03:53:30Z ERR Connection terminated error="failed to dial to edge with quic: timeout: no recent network activity" connIndex=1
2022-08-11T03:53:30Z WRN If this log occurs persistently, and cloudflared is unable to connect to Cloudflare Network with `quic` protocol, then most likely your machine/network is getting its egress UDP to port 7844 (or others) blocked or dropped. Make sure to allow egress connectivity as per https://developers.cloudflare.com/cloudflare-one/connections/connect-apps/configuration/ports-and-ips/
If you are using private routing to this Tunnel, then UDP (and Private DNS Resolution) will not work unless your cloudflared can connect with Cloudflare Network with `quic`. connIndex=3 ip=198.41.192.47
2022-08-11T03:53:30Z INF Switching to fallback protocol http2 connIndex=3 ip=198.41.192.47
2022-08-11T03:53:30Z ERR Connection terminated error="failed to dial to edge with quic: timeout: no recent network activity" connIndex=3
[05:53:34] WARNING: Connection unavailable, rechecking in 5 seconds.
[05:53:34] WARNING: Connection attempt 15/24 before restart.
2022-08-11T03:53:38Z ERR Unable to establish connection with Cloudflare edge error="TLS handshake with edge error: read tcp 172.30.33.5:57668->198.41.200.33:7844: i/o timeout" connIndex=0 ip=198.41.200.33
2022-08-11T03:53:38Z ERR Serve tunnel error error="TLS handshake with edge error: read tcp 172.30.33.5:57668->198.41.200.33:7844: i/o timeout" connIndex=0 ip=198.41.200.33
[...]

PR #143, opened 2022-08-12
Add watchdog URL

elcajon:
# Proposed Changes

I'm not sure if this works, the add-on developer documentation doesn't give that much at this point. But it's worth a try.

## Related Issues

#141 


brenner-tobias (approved):
Not sure either but let's give it a try

PR #151, opened 2022-08-15
Watchdog - fix metrics server binding

elcajon:
# Proposed Changes

Should provide Watchdog the possibility to contact metrics server for healthchecks issued from HA supervisor. 
This solution is working locally for ~60 mins now. 

If this isn't working we need to remove the watchdog string from add-on config.

## Related Issues


brenner-tobias (approved):
this is strange, looking at the Cloudflared documentation it should work with "localhost:".
Let's give it a try and see, thanks for fixing

elcajon:
It works from cloudflare perspective, but watchdog is called from supervisor container and therefore cant reach metrics endpoint which is bound to localhost inside cf docker.

brenner-tobias:
ah yeah, makes sense. Thanks for clarifying

PR #153, opened 2022-08-15
Alter loglevel of unavailable connection dependent on counter

brenner-tobias:
# Proposed Changes

Alter loglevel of healthcheck if connection was not successful dependent on counter. Only starting from the third unsuccessful connection, it will be a warning. Before that, it will be a debug message.
This mainly helps reduce confusion for users when starting the add-on, where one warning is created most of the time.


brenner-tobias:
@elcajon please let me know what you think of this and suggest any improvements you have.

elcajon (approved):

elcajon:
I like that, I think that will eliminate some unnecessary questions. 👍

PR #156, opened 2022-08-16
Optimize Cloudflare_release Workflow

brenner-tobias:
# Proposed Changes

> Change reviewers to assignies for automatically created PR from cloudflare_release workflow. That way, (hopefully) not two reviews are needed, but all relevant parties are still notified.

> Add label "dependencies" to PR


brenner-tobias:
@elcajon please let me know if you have any other suggestions / better ideas

elcajon:
Two reviews should not be necessary due to the branch protection settings. For the last PR I did not submit a review. 
What we need to do because of the automatic generated PR is to close the PR and reopen it. After that, the CI will start and can be merged after one approval.
Nevertheless I think the change doesn't harm, so lets go for it. 👍

elcajon (approved):

PR #164, opened 2022-08-21
Check active network connection first

elcajon:
# Proposed Changes

This should fix all of the currently reported internet outage errors.

Given the following conditions:
- Connection outage while add-on connected --> implemented health check which restarts add-on
- Connection outage during start up phase --> this PR adds a check if host is connected to internet

## Related Issues


elcajon:
This is not working as expected, I haven't got the time right now to dig deeper but will investigate the next days.

elcajon:
Alright, I changed logic to simply ping cloudflare at 1.1.1.1 to check if internet is available. This should do the trick at add-on startup to prevent errors from cloudflare daemon as well during health checks to wait for network to be restored before checking cloudflare health again.

brenner-tobias (approved):

PR #171, opened 2022-09-02
Remove add-on healthcheck

elcajon:
# Proposed Changes

## Related Issues


elcajon:
@brenner-tobias  Please don't release next version, yet. I want to update some other dependencies first.

cloudflared/config.yaml
```diff
@@ -9,7 +9,6 @@ codenotary: dev@brenner.tech
 init: false
 hassio_api: true
 hassio_role: homeassistant
-watchdog: "http://[HOST]:36500/healthcheck"
```
brenner-tobias, line 12: Do you think that we should completely remove watchdog for the add-on?
I think it would be great to have some sort of "auto-restart" if it crashes.
I am fine with removing it for now and thinking of another solution, just wanted to mention it here.

brenner-tobias (approved):
Fine for me. Will not release a new version though. Again thinking about leaving watchdog to some extend or re-adding it at another stage

cloudflared/config.yaml
elcajon: I would suggest moving this (as recommended) to Dockerfile as HEALTHCHECK statement, but this needs to be tested first.

The current logic without watchdog statement inside config.yaml will (when Watchdog switch is set to on) just restart the add-on when it crashes with exit code != 0

PR #173, opened 2022-09-02
Update cont-init.d scripts from base to s6 v3 with system health check

elcajon:
# Proposed Changes

## Related Issues


elcajon:
That's the last one for now, feel free to release draft after some tests

brenner-tobias (commented):
Thank you very much. We did not need to manually add the banner and the log level before, right? So why do we need to do this now?

cloudflared/rootfs/etc/s6-overlay/scripts/00-banner.sh
```diff
@@ -0,0 +1,53 @@
...
+    bashio::log.magenta
+fi
+
+if bashio::var.false "$(bashio::supervisor.healthy)" ; then
+    bashio::log.magenta
+    bashio::log.magenta " System is unhealthy."
+    bashio::log.magenta " Errors with this add-on may occur."
+    bashio::log.magenta " Before asking for support fix your system health errors."
+    bashio::log.magenta
+fi
```
elcajon, line 43: Because of this, this is not in the base image and we've got a lot of issues coming from systems which aren't officially supported or unhealthy.

brenner-tobias: Thanks for that, totally makes sense. Do you think it makes sense to add this as a PR for the base image?

PR #181, opened 2022-09-20
Move "docs" in "cloudflare" to include in add-on repository

brenner-tobias:
# Proposed Changes

> Move "docs" folder into "cloudflare" in order to include it in the ha-addons repository.
> see https://github.com/brenner-tobias/ha-addons/issues/5


elcajon (approved):

elcajon:
I am not sure if this folder will be copied but let's give it a try, otherwise we need to change the referring links to use the full domain.

PR #184, opened 2022-09-21
Tidy add-on description

AlecRust:
"Use a Cloudflared tunnel" reads wrong to me. It's a Cloudflare Tunnel, and it's achieved via the Cloudflare Tunnel client (formerly Argo Tunnel) called `cloudflared`.

Perhaps the finer details of this don't need to be included in the description, but "Cloudflared tunnel" stood out as wrong to me.

### DIFF
--- a/cloudflared/DOCS.md
+++ b/cloudflared/DOCS.md
@@ -52,7 +52,7 @@ still persists.
 
 Workaround:
 
-1. Create a new Cloudflare Account and invite it to your Cloudflare account
+1. Create a new Cloudflare account and invite it to your Cloudflare account
    that manages your Domain:\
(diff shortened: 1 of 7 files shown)


elcajon:
I would agree on that, what do you think @brenner-tobias? 
Do we also need to review/adapt the rest of the docs etc.?

brenner-tobias:
Agreed, let's search through the documentation so it is consistent everywhere. Can you @AlecRust have a look in your branch and update the PR accordingly? Also, please make sure that all the checks run through, then we can merge.

AlecRust:
Sure, I've pushed a more thorough tidying.

elcajon (approved):

PR #189, opened 2022-09-23
Update docs with mentions of security

AlecRust:
It seems common to set up WAF rules and utilise Cloudflare Access as an interstitial page before redirecting to Home Assistant.

This updates the docs to include mention of these approaches, with links.

I've also moved this under the main setup instructions. First time I read it I assumed "Securing access to the Cloudflare account" was part of initial setup, but it actually seems more of a troubleshooting section.


brenner-tobias (approved):
Thanks a lot for this PR

PR #203, opened 2022-10-03
Make external_hostname and tunnel_name optional

brenner-tobias:
# Proposed Changes

- Enable usage without using _external_host_ (for _custom_config_, _additional_hosts_, _catch_all_service_ or _nginx_proxy_manager_
- Make _tunnel_name_ optional with default "homeassistant"

## Related Issues

#190

### DIFF
--- a/cloudflared/rootfs/etc/s6-overlay/scripts/10-cloudflared-config.sh
+++ b/cloudflared/rootfs/etc/s6-overlay/scripts/10-cloudflared-config.sh
@@ -29,14 +29,15 @@ checkConfig() {
         bashio::exit.ok "Running with Custom-Config and skipping further validations"
     fi
 
-    # Check if 'external_hostname' is a non-empty string
-    if bashio::config.is_empty 'external_hostname' ; then
-        bashio::exit.nok "'external_hostname' is empty, please enter a valid String"
-    fi
-
-    # Check if 'external_hostname' includes a valid hostname
-    if ! [[ $(bashio::config 'external_hostname') =~ ${validHostnameRegex} ]] ; then
-        bashio::exit.nok "'$(bashio::config 'external_hostname')' is not a valid hostname. Please make sure not to include the protocol (e.g. 'https://') nor the port (e.g. ':8123') in the 'external_hostname'."
+    # Check if 'external_hostname' includes a valid hostname 
+    if bashio::config.has_value 'external_hostname' ; then
+        if ! [[ $(bashio::config 'external_hostname') =~ ${validHostnameRegex} ]] ; then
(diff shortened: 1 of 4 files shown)


brenner-tobias:
I currently cannot test this properly, so please feel free to suggest any documentation changes or code fixes

cloudflared/rootfs/etc/s6-overlay/scripts/10-cloudflared-config.sh
```diff
@@ -256,10 +259,12 @@ createConfig() {
 createDNS() {
```
brenner-tobias, line 259: When looking at this method, I figured we might as well rename 'createCustomDNS' to 'createDNS' and use it all the time. Whether it is a custom_config or a config created by the add-on should not matter, right? @elcajon

elcajon: The difference is we use json for automatically generated config and YAML for custom config. We would need to streamline this first or simply combine both functions which wouldn't really reduce complexity.

elcajon:
Technically, the changes you made work fine. 👍🏻

I think we should adjust the documentation to facilitate the configuration of the add-on. 

The documentation has to make it clear that the choice between local and remote managed tunnel must be made first.

- For using the remote managed tunnel `tunnel_token` needs to be set
- For using the local managed tunnel external_hostname and/or additional_hosts need to be set.
- All other parameters can be used in addition for the local managed tunnel.

In this context (maybe as additional PR), I suggest to remove tunnel_name from the default configuration and set it to the default value "homeassistant" in case of an empty value. Thus we can set the parameter optionally.

cloudflared/rootfs/etc/s6-overlay/scripts/10-cloudflared-config.sh
brenner-tobias: Yeah true, I forgot about that. Let's leave it as it was

elcajon (approved):

elcajon:
Looks good, I would say lets merge this and add documentation in a separate PR

brenner-tobias:
> Looks good, I would say lets merge this and add documentation in a separate PR

sorry, I missed your comment and had another commit lined up. Can you have another quick look @elcajon?

elcajon:
Awesome, looks good. 👍🏻

elcajon (approved):

elcajon:
I think we could post a new release after this PR and reference the survey again. In parallel we could already remove the warp functionality, the feedback is relatively low on this configuration option.

AlecRust:
You guys are fast! Thank you for this change 🙏 

Looks like you've already sorted the docs, but when this goes in I'll follow it with a minor docs tidy PR.

PR #207, opened 2022-10-07
Change order to reflect the default options positions

brenner-tobias:
# Proposed Changes

- Change the order of the options to have the ones with default values at the very top. That way, they are not "jumping" if you toggle the "show optional options" switch.


brenner-tobias:
I mainly just want a new release for the edge version

elcajon (approved):

PR #208, opened 2022-10-07
Tidy docs

AlecRust:
- Docs talk about deciding between "local or managed tunnels" then the next section is instructions for "remote tunnels" which may cause slight confusion.
- "Installation" section in `DOCS.md` doesn't explain how or when to actually install the add-on.
- Remote tunnel instructions are listed first, but local tunnel is the recommendation
- Various other minor improvements


brenner-tobias (approved):
Awesome, thank you very much.

PR #210, opened 2022-10-09
Remove warp, deprecate custom config

brenner-tobias:
# Proposed Changes

- Remove warp_* options
- Deprecate custom_config

## Related Issues

#191


brenner-tobias:
Let's keep this PR open for another few weeks and only merge it then

brenner-tobias:
I reconsidered and think the following is better: Deprecate custom_config now as well (no breaking change). Remove both warp and custom config in a couple of weeks together with one breaking change.

PR #211, opened 2022-10-09
Deprecate custom_config

brenner-tobias:
# Proposed Changes

- Deprecate custom_config

## Related Issues

#191


brenner-tobias (commented):
I suggest we deprecate custom_config now as well and then remove both warp_* and custom_config in one breaking release.

elcajon (approved):

PR #212, opened 2022-10-09
Deprecate 'data_folder'

brenner-tobias:
# Proposed Changes

> Deprecate 'data_folder'

## Related Issues

#191

### DIFF
--- a/cloudflared/DOCS.md
+++ b/cloudflared/DOCS.md
@@ -74,7 +74,7 @@ advanced configurations can be achieved using the remote tunnel setup.
 - [`additional_hosts`](#option-additional_hosts)
 - [`catch_all_service`](#option-catch_all_service)
 - [`nginx_proxy_manager`](#option-nginx_proxy_manager)
-- [`data_folder`](#option-data_folder)
+- [`data_folder (Deprecated)`](#option-data_folder)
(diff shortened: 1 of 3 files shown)


brenner-tobias:
I did consider keeping 'data_folder' but it is really not useful without a custom_config from my perspective.
The only thing I am unsure of: Should we include a "migrate_back" flag with a function to migrate the tunnel and certificate files back to the default location or just state that the user should do it?

elcajon:
I would suggest to automatically copy the cert.pem and tunnel.json file from the configured data_path back to /data and showing a warning message. This way should be non disruptive as many users may not know how to access /data paths of add-ons.

elcajon:
Maybe this  "migration" could also be added before removing the configuration options (breaking update), then we just need to reset data_path back to default/blank. This would give users time to raise any errors related to this file moving process.

brenner-tobias:
yeah let's do it that way. I will make a suggestion here

brenner-tobias:
Added a rather harsh solution: If someone is only using 'data_folder' without 'custom_config', the files will be migrated automatically and the option will be removed. Otherwise, the "normal" warning from 'custom_config' will be shown.

elcajon (approved):
Looks good 👍🏻

PR #213, opened 2022-10-09
set data path for default migration

brenner-tobias:
# Proposed Changes

Fix that sets data_path correctly for migration to default location

## Related Issues

### DIFF
--- a/cloudflared/rootfs/etc/s6-overlay/scripts/10-cloudflared-config.sh
+++ b/cloudflared/rootfs/etc/s6-overlay/scripts/10-cloudflared-config.sh
@@ -438,11 +438,13 @@ main() {
             bashio::log.warning "You are using the data_folder option"
             bashio::log.warning "Please note that this option is deprecated and will be removed soon."
             bashio::log.warning "We will auto-migrate your files to the default location '/data'"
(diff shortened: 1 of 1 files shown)


cloudflared/rootfs/etc/s6-overlay/scripts/10-cloudflared-config.sh
```diff
@@ -438,11 +438,13 @@ main() {
             bashio::log.warning "You are using the data_folder option"
             bashio::log.warning "Please note that this option is deprecated and will be removed soon."
             bashio::log.warning "We will auto-migrate your files to the default location '/data'"
+            data_path="/$(bashio::config 'data_folder')/cloudflared"
             migrateFilesToDefault
             bashio::log.warning "Sucessfully migrated your files to the default location"
             bashio::log.warning "Removing the 'data_folder' option"
             bashio::addon.option 'data_folder'
             bashio::log.warning "Starting add-on with default location"
+            data_path="/data"
```
elcajon, line 447: This is not going to work. We check for custom config again later (line 466) and function hasCustomConfig will exit nok if config.yml doesn't exist.

I would suggest to remove this line during deprecation state to be able to handle custom config files correctly.

PR #214, opened 2022-10-09
Add creation of dir if not existing

brenner-tobias:
# Proposed Changes

- Fix creation of custom folder if not existing


brenner-tobias:
will be addressed in #215

PR #215, opened 2022-10-09
Update 10-cloudflared-config.sh

elcajon:
# Proposed Changes

Bugfix for data_path option.
Also reconsidering not to remove value for data_path so custom_config is still working after add-on restart.

## Related Issues

### DIFF
--- a/cloudflared/rootfs/etc/s6-overlay/scripts/10-cloudflared-config.sh
+++ b/cloudflared/rootfs/etc/s6-overlay/scripts/10-cloudflared-config.sh
@@ -429,8 +429,8 @@ main() {
 
     # Check for custom data path
     if bashio::config.has_value 'data_folder'; then
+        data_path="/$(bashio::config 'data_folder')/cloudflared"
         if bashio::config.true 'custom_config' ; then
-            data_path="/$(bashio::config 'data_folder')/cloudflared"
             bashio::log.info "Data path set to ${data_path}"
             mkdir -p "${data_path}"
             migrateFiles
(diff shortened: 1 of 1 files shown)


brenner-tobias:
Thanks for that @elcajon, I basically just created the same fix in #214, but let's use this branch.
Please have a look at my comment

cloudflared/rootfs/etc/s6-overlay/scripts/10-cloudflared-config.sh
```diff
@@ -429,22 +429,19 @@ main() {
...
             migrateFiles
         else
             bashio::log.warning "You are using the data_folder option"
             bashio::log.warning "Please note that this option is deprecated and will be removed soon."
             bashio::log.warning "We will auto-migrate your files to the default location '/data'"
-            data_path="/$(bashio::config 'data_folder')/cloudflared"
             migrateFilesToDefault
             bashio::log.warning "Sucessfully migrated your files to the default location"
             bashio::log.warning "Removing the 'data_folder' option"
-            bashio::addon.option 'data_folder'
```
brenner-tobias, line 445: So my approach was: We are here in the condition that the user is only using data_path but not custom_config. So we can simply just migrate back to the default and then also remove the option.
I do not think that this is an problem, since the add-on will still work after a restart and also with custom_config.

The other possibility here would be to not remove the option and also not set data_path (as suggested by you), but then we also have to remove the two logg entries and remember to include a call to remove the option when deleting in the next release if present.

brenner-tobias (changes requested):

elcajon:
Never mind this PR. Everything should be fine just merging your PR.
I have misinterpreted something.

cloudflared/rootfs/etc/s6-overlay/scripts/10-cloudflared-config.sh
elcajon: That's exactly what I just overlooked, so it should work like that. Sorry for the back and forth.

brenner-tobias: all good, thanks a lot for checking. I merged my PR, should be fine now

PR #217, opened 2022-10-11
Remove custom config, data_folder and warp

brenner-tobias:
# Proposed Changes

- Remove deprecated options 'custom_config', 'data_folder' and 'warp_*'
- Remove 'yq' from Docker container, since it is not used anymore

## Related Issues

#191


brenner-tobias:
Let's not merge this for another week or so @elcajon

cloudflared/rootfs/etc/s6-overlay/scripts/10-cloudflared-config.sh
```diff
@@ -7,6 +7,58 @@
 # given hostname(s)
 # ==============================================================================
 
+# ------------------------------------------------------------------------------
+# Checks for legacy options and removes them
+# ------------------------------------------------------------------------------
+checkLegacyOptions() {
```
brenner-tobias, line 13: @elcajon I would love to hear your thoughts on checking this here, including keeping the migration from the 'data_folder'. The alternative would be just not to care about the legacy options being present in the config and not start if the config does not meet our "normal" checks

elcajon: I would say that this option is more user friendly then just failing with an error. Let's start with this for some minor or major releases before removing the functionality again.

elcajon (approved):

PR #228, opened 2022-10-23
Tidy docs

AlecRust:
Some tidying up, including fixing `external_hostname` not having a documentation section which is potentially confusing.


brenner-tobias (approved):
Thanks a lot

PR #266, opened 2022-12-12
⬆️ Update base image to v13.0.0

elcajon:
# Proposed Changes

- Base image v13.0.0 includes new s6 cont-init-scripts (instead of legacy ones) --> I therefore removed the now unnecessary files
- Because HA log level is no longer needed, I integrated Cloudflared log into cloudflared-config init script.
- Adaption of S6 log sensitivity to remove not needed log output.
- Removed services for (not needed) legacy init scripts (fixattr, cont-init.d, services.d)

## Related Issues


cloudflared/rootfs/build.sh
```diff
@@ -40,3 +40,8 @@ chmod +x /usr/bin/cloudflared
 
 # Remove legacy cont-init.d services
 rm -rf /etc/cont-init.d
+
+# Remove s-6 legacy/deprecated (and not needed) services
```
brenner-tobias, line 44: Can you elaborate on why we are doing this and if this will become irrelevant at some point?

cloudflared/Dockerfile
```diff
@@ -1,10 +1,13 @@
-ARG BUILD_FROM=ghcr.io/hassio-addons/base/amd64:12.2.7
+ARG BUILD_FROM=ghcr.io/hassio-addons/base/amd64:13.0.0
 # hadolint ignore=DL3006
 FROM ${BUILD_FROM}
 
 # Set shell
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 
+# Set S6 verbosity level
+ENV S6_VERBOSITY=1
```
brenner-tobias, line 9: Do we need this? What is the default?

brenner-tobias:
Thanks a lot @elcajon

brenner-tobias:
@elcajon after re-reading your proposed changes, all my questions are closed. Thanks again

brenner-tobias (approved):

cloudflared/rootfs/build.sh
elcajon: This is only necessary if we would run scripts from custom-cont-init.d or services.d which is in maintenance only mode since s6 version 3. We don't need it as we migrated to s6 v3 init scripts and services. I guess we can save som system resources not starting the daemons (but not sure tbh)

cloudflared/Dockerfile
elcajon: default would be "2" which outputs lines like these (info, warning & error):

```
s6-rc: info: service s6rc-oneshot-runner: starting
s6-rc: info: service s6rc-oneshot-runner successfully started
s6-rc: info: service fix-attrs: starting
s6-rc: info: service init-banner: starting
s6-rc: info: service fix-attrs successfully started
s6-rc: info: service legacy-cont-init: starting
s6-rc: info: service legacy-cont-init successfully started
```

value "1" should suppress "info" messages and only logs "warning" and "error" messages. 
I guess this will reduce logs provided for issues/bugs at least a bit.

PR #273, opened 2022-12-20
Update CI

elcajon:
# Proposed Changes

## ci.yaml
- update versions
- update set-output to env logic (set-output is deprecated)
- change Codenotary logic as the old Github action uses Github API --> rate limit problems

## deploy.yaml
- update versions
- update set-output to env logic (set-output is deprecated)
- change Codenotary logic as the old Github action uses Github API --> rate limit problems

## cloudflare_release.yaml --> release_updater.yaml
- update set-output to env logic (set-output is deprecated)
- change release updater logic as the old Github action uses Github API --> rate limit problems

I also updated the workflow to the latest version which I use to keep my Github binaries up to date. 
This version brings two new features:
- generic update approach to update more than one Github release (not really needed as we only have cloudflare/cloudflared in use)
- completely automate the update process 
-- create PR (as different User [PAT] or authorized Github App --> triggers workflow run for PR without manually closing & reopening)
-- set PR to auto merge if all status checks pass (CI + Review)
-- add approval review

I commented out all the new optional features as PAT (from second account) or Github app needs to be set up first. I also don't know if you want to have this level of automation 😉


brenner-tobias (commented):
@elcajon thanks a lot, this is awesome. Just to clarify before approving: The way it is setup now is that we will remain in the same process as now: There will be a PR for the Cloudflare Releases and we have to approve and merge, right?

elcajon:
I did a test in a separate repository with the PR version for release updates and everything went smooth. As it is not possible to run workflows from PRs, we should trigger a run after merging. However, I do not expect any major problems... I am also available tonight to troubleshoot any issues that arise if needed.

brenner-tobias (approved):

elcajon:
#275 👍🏻

brenner-tobias:
@elcajon you did close and re-open the PR, I thought with the new system this was not necessary anymore to trigger the Actions?

elcajon:
This would only work if we change the token to Github App or PAT. The current implementation behaves exactly as the previous one.

brenner-tobias:
ah got it, sorry for missing that. Let's change it, I am happy with both approaches (for a PAT we could use my [brenner-bot](https://github.com/brenner-bot)). Which approach do you recommend?

elcajon:
I would also recommend to use a PAT from your bot account. I guess you need to add the account as contributor and save a PAT as secret but then it should work.

PR #307, opened 2023-01-26
Remove Freenom reference, add wiki link

brenner-tobias:
# Proposed Changes

Since Freenom seems not to be working anymore, we should remove the references to freenom and link to our wiki instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,9 +24,9 @@ instance and other services to the Internet without opening ports on your router
 Additionally, you can utilize Cloudflare Zero Trust to further secure your
 connection.
 
-**To use this add-on, you need a domain name (e.g. example.com) that is using
-Cloudflare for its DNS entries. If you do not have one, you can get one for
-free at [Freenom][freenom] following [this article][domainarticle].**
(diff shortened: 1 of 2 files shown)


AlecRust (approved):

lin-ycv:
Can I get a clarification on how the new update affects freenom usage?
I'm using freenom and it works fine, if I update to the new release, would it break?

dankarization:
Also confirming that existing freenom domains are working fine

adam-cobb:
Slightly concerned by this update as I have two instances of HA both using freenom domains in Cloudflare, both working fine currently as the others above. Will hold of any update until we can get some clarification. Thanks.

brenner-tobias:
As you can see in the release notes and also in the code, the removal of Freenom is only a documentation topic, though I agree that this could have been phrased even more clearly. There are numerous reports of people having major issues with Freenom domains, so I removed the part of the documentation that suggest to get a domain from Freenom if you don't have one. I replaced it with the suggestion to buy a cheap one from Namecheap, GoDaddy or any other registar.

So if everything is working now for you, it will continue to work, so you can update without expecting any issues.

PR #346, opened 2023-03-20
Remove uppercase chars from valid hostnames

### LINKED ISSUE #327: Every couple weeks I get Error 1003
### The problem

Cloudflared works great then every couple weeks gets this error and I have to log into Cloudflare, delete the DNS CName entry, and restart Cloudflared add-on.
  

Screenshot 2023-02-20 145539
Screenshot 2023-02-20 145512

4.0.10

_No response_

Home Assistant OS

### Add-on YAML Configuration

```yaml
additional_hosts: []
external_hostname: <Redacted> // it is just my personal domain name
```

_No response_

Everything works perfectly, then every week or couple weeks, it gets the error shown in the screenshots above.  It's like the add-on is trying to make a new tunnel when there's already one there that it made before.  I'll note that I do not do anything else with Cloudflare or this domain except use it as documented for the add-on.

### Additional information

_No response_

### LINKED ISSUE #342: Need to delete cloudflare DNSs everytime before i restart the addon
### The problem

I had the same problem problem described in issue #259 tonight.  Long story short, after installing and configuring the add-on, the tunnel is created successfully.  However at the next restart it fails to create the DNS entry with this message in the log:

[09:38:09] INFO: Creating DNS entry ###.###.tk...
Failed to add route: code: 1003, reason: An A, AAAA, or CNAME record with that host already exists.
[09:38:11] FATAL: Failed to create DNS entry ###.###.tk.

The problem is when the "external_hostname" is written in CAPITALS, you will have the issue, because the CNAME DNS entry in the Cloudflare Portal is stored in lowercase. 

At the next load, when the hostname is compared with existing DNS entries, the comparaison fails because the verification is probably case sensitive, then it tries to recreate the entry and fails because it already exists.

4.0.12

_No response_

Home Assistant Supervised

### Add-on YAML Configuration

_No response_

```txt
[09:38:09] INFO: Creating DNS entry ###.###.tk...
Failed to add route: code: 1003, reason: An A, AAAA, or CNAME record with that host already exists.
[09:38:11] FATAL: Failed to create DNS entry ###.###.tk.
```

In an existing installation, if the hostname is written in capitals (ex. HOST.domain.com), the add-on will fail to load.

### Additional information

_No response_

brenner-tobias:
# Proposed Changes

Not allow uppercase characters in hostnames (external_hostname and in additional_hosts)

## Related Issues

Closes #342 
Closes #327

### DIFF
--- a/cloudflared/rootfs/etc/s6-overlay/scripts/cloudflared-config.sh
+++ b/cloudflared/rootfs/etc/s6-overlay/scripts/cloudflared-config.sh
@@ -26,7 +26,7 @@ checkConfig() {
     # Check if 'external_hostname' includes a valid hostname
     if bashio::config.has_value 'external_hostname' ; then
         if ! [[ $(bashio::config 'external_hostname') =~ ${validHostnameRegex} ]] ; then
-            bashio::exit.nok "'$(bashio::config 'external_hostname')' is not a valid hostname. Please make sure not to include the protocol (e.g. 'https://') nor the port (e.g. ':8123') in the 'external_hostname'."
+            bashio::exit.nok "'$(bashio::config 'external_hostname')' is not a valid hostname. Please make sure not to include the protocol (e.g. 'https://') nor the port (e.g. ':8123') and only use lowercase characters in the 'external_hostname'."
         fi
     fi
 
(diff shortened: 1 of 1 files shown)


cloudflared/rootfs/etc/s6-overlay/scripts/cloudflared-config.sh
```diff
@@ -14,7 +14,7 @@ checkConfig() {
     bashio::log.trace "${FUNCNAME[0]}"
     bashio::log.info "Checking add-on config..."
 
-    local validHostnameRegex="^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])$"
+    local validHostnameRegex="^(([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])\.)*([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])$"
```
elcajon, line 17: If the regex is changed, the error message after an unsuccessful check would probably have to be changed as well, right? Alternatively, it would also be possible to simply convert the string by using `bashio::string.lower`

brenner-tobias: Agreed, I did include the message in my last commit. I also thought about simply changing it, but I would only do this if we are also able to actually change the config entry itself and not just write lowercase to Cloudflare. To do that is not super simple, especially for the additional_hosts array.
So I think it is easier and also clearer to just print the error message and have the users change it. Happy to hear your suggestion though

elcajon (approved):

PR #353, opened 2023-03-30
Add option in documentation for wildcard DNS record

gerard33:
# Proposed Changes

Add option in documentation for wildcard DNS record as described [here](https://blog.cloudflare.com/wildcard-proxy-for-everyone/).

### DIFF
--- a/cloudflared/DOCS.md
+++ b/cloudflared/DOCS.md
@@ -172,7 +172,9 @@ Any other incoming traffic will be routed to the defined service._
 In order to route hostnames through the tunnel, you have to create individual
 CNAME records in Cloudflare for all of them, pointing to your `external_hostname`
 or directly to the tunnel URL that you can get from the CNAME entry of
-`external_hostname`.
+`external_hostname`.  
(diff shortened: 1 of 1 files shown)


elcajon:
In my opinion, the documentation is rather misleading. CNAME records are created automatically by the add-on and do not need to be created manually. Also, using wildcard in combination with additional_hosts is difficult, because there can be overlaps.
I think this is a use case we should not promote via the CLI version of the add-on (local managed tunnel). As in my opinion further testing is needed to ensure the function.

gerard33:
The information I added is meant for the `catch_all_service` and `nginx_proxy_manager` where you have to manually add CNAME records in Cloudflare. This can be skipped when using the wildcard DNS record.
And if you enter an individual CNAME record in Cloudflare than that overrules the wildcard record. I have tested that.

elcajon:
Understood, thanks for the clarification. For me the change fits. Thanks for the contribution. 👍🏼

brenner-tobias (approved):
Thanks a lot, great addition to the documentation

PR #357, opened 2023-04-02
Add Post-Quantum to tunnel

brenner-tobias:
# Proposed Changes

Add --post-quantum flag to all tunnel connections


elcajon:
From the [blog](https://blog.cloudflare.com/post-quantum-tunnel/) entry:

_When the --post-quantum flag is given, cloudflared restricts itself to the QUIC transport for the tunnel connection to our network and will only allow the post-quantum_ 

Can this cause problems in situations where users cannot connect via QUIC due to their ISP and therefore a fallback to http2 is necessary?

brenner-tobias:
very good point. Maybe we should introduce a flag "disable_post-quantum" or something like that (or maybe simply a positive flag "set_post-quantum" that is enabled by default) and explain in the documentation to set / unset this, if there are problems with QUIC?

elcajon:
To avoid unnecessary error messages I would rather be for an optional opt-in via optional add-on option

cloudflared/rootfs/etc/s6-overlay/s6-rc.d/cloudflared/run
```diff
@@ -7,9 +7,13 @@ declare default_config=/tmp/config.json
 
 if bashio::config.has_value 'tunnel_token' ; then
```
brenner-tobias, line 8: @elcajon since we now check two config values, there might be a better way to "construct" the cloudflared excec string (since they also always share a lot). Any suggestions?

elcajon:
I have added a proposal to restructure the service call and tested it successfully for local and remote managed tunnels.

elcajon (dismissed):

cloudflared/rootfs/etc/s6-overlay/s6-rc.d/cloudflared/run
```diff
@@ -3,23 +3,39 @@
 # Home Assistant Add-on: Cloudflared
 # Runs the Cloudflare Tunnel for Home Assistant
 # ==============================================================================
-declare default_config=/tmp/config.json
+declare config_file="/tmp/config.json"
+declare certificate="/data/cert.pem"
+declare -a cloudflared_options
```
brenner-tobias, line 8: This is great. Not sure though if we need the separation of "cloudflared_options" and "tunnel_options".

Looking at the [documentation](https://developers.cloudflare.com/cloudflare-one/connections/connect-apps/install-and-setup/tunnel-guide/local/local-management/arguments/), you can add the "--no-autoupdate" also after "tunnel". I thin the same should be valid for the --post-quantum.

What do you think?

elcajon: You are right, I have changed it. That simplifies the whole thing enormously. 😉

elcajon (approved):

brenner-tobias (commented):
Approved (but cannot properly approve it  in GitHub, since this is my PR). @elcajon feel free to approve as well, so we can merge

PR #362, opened 2023-04-03
Revise the documentation

elcajon:
# Proposed Changes

## Related Issues

### DIFF
--- a/cloudflared/DOCS.md
+++ b/cloudflared/DOCS.md
@@ -85,12 +85,12 @@ advanced config can be achieved using the remote tunnel setup.
 Example add-on configuration:
 
 ```yaml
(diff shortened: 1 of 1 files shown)


elcajon:
Just a thought, but what do you say we recommend the remote managed tunnel and classify the local managed tunnel as advanced?

brenner-tobias:
> Just a thought, but what do you say we recommend the remote managed tunnel and classify the local managed tunnel as advanced?

Not sure, since I guess for most users it is still easiest to just use it for HA with the external hostname and are done. This way, they do not have to create the Zero Trust Account (which requires payment methods, though they are not charging anything).
I suggest to leave it for now and reconsider regularly if they introduce new features at Cloudflared

PR #433, opened 2023-07-11
Add docker health check based on cloudflared tunnel status

elcajon:
# Proposed Changes

WIP

## Related Issues

### DIFF
--- a/cloudflared/Dockerfile
+++ b/cloudflared/Dockerfile
@@ -18,6 +18,10 @@ COPY rootfs /
 # Run the script that installes cloudflared
 RUN chmod a+x /build.sh && /build.sh "${BUILD_ARCH}" "${CLOUDFLARED_VERSION}"
 
+# Health check
+HEALTHCHECK \
+    CMD curl --fail 127.0.0.1:36500/ready | jq -e 'select(.status == 200).status' || exit 1
+
 # Build arguments
 ARG BUILD_ARCH
 ARG BUILD_DATE


brenner-tobias:
Thanks for this. Indeed the initial authentification is a problem. What will failing the healthcheck trigger?
Maybe we can work with the flag "retries" (which should be defaulting to 3?) to increase the actual time until the check fails?

elcajon:
https://docs.docker.com/engine/reference/builder/#healthcheck

```
When a container has a healthcheck specified, it has a health status in addition to its normal status. 
This status is initially starting. Whenever a health check passes,
it becomes healthy (whatever state it was previously in). 
After a certain number of consecutive failures, it becomes unhealthy.
```

Start period for HA OS/Supervisor is 120s
```
start period provides initialization time for containers that need time to bootstrap. 
Probe failure during that period will not be counted towards the maximum number of retries.
However, if a health check succeeds during the start period, the container is considered started
and all consecutive failures will be counted towards the maximum number of retries.
```

I'm afraid that we don't have many options here to change this default behavior. 
I am currently also not sure if the benefits of a healthcheck outweigh the negative aspects.

brenner-tobias:
If we look at the possible configurations, we should be able to stretch it, right?

- interval - DURATION (default: 30s)
- timeout - DURATION (default: 30s)
- start-period - DURATION (default: 0s)
- retries - DURATION (default: 3)

So if we cannot change the start-period to more than 120, we can set change the interval, timeout and retires.

My understanding is, that even in our current setting, the timing should be the following:

1. Wait for 120 sec
2. Check, get immediate fail, wait 30 sec, repeat for 3 times

So in total, we should be at 210 seconds (120 + 3 * 30), right? If this is correct, I think changing the interval to e.g. 40s should do the trick with 240s, so 4 Minutes, which should be fine, otherwise you might run into timeout issues with Cloudflare anyways. Of course this also means, that the healthcheck only fails after 120 seconds once it was running properly.

It is really strange that a value higher than 120 for the start-period is not working. Did you find any information about why this is not working? It would indeed be best for this number to just be at 240 and that's it.

elcajon:
As I understand it, we are in "starting" status until the first healthcheck is successful. Unfortunately we can't get out of this status until the login is done and I don't know how we can force a successful healthcheck during the login phase without Metrics Server.
This is problematic because the add-on is automatically stopped (or restarted, if watchdog is enabled) by the supervisor after 120 seconds in "starting" status.

elcajon:
> Did you find any information about why this is not working? It would indeed be best for this number to just be at 240 and that's it.

No, that also surprised me when I tested the function in general. There was only one entry in the supervisor log and the add-on was suddenly stopped again

brenner-tobias:
Got it, I missed the piece that we had to first pass it once. To be frank I am uncertain here. On one hand, two minutes is not super short, but even if you work pretty fast, you have to find your Cloudflare Login and also select the right domain during that time. Additionally, the connection needs to be established, so I feel we are introducing a lot of struggle for the initial set-up (if it is not via the Cloudflare Dashboard).

So my gut feeling is tho keep looking for a solution (maybe raise an issue at HA regarding those 120 sec?) and not merge it until then since we will introduce more problems than have a benefit.

elcajon:
I agree with you, there seems to be movement at the moment anyway. Something will also change soon with regard to Codenotary. I'll keep an eye on that.

BTW - Strange, I just started another test and now the add-on has been running for 10 minutes without being stopped at all...

Nevermind I am closing this for the moment. 😉

elcajon:
Update:

appears in the supervisor log but independent from watchdog
the add-on stops after 10 minutes because of an internal Cloudflare timeout
```
WARNING (MainThread) [supervisor.addons.addon] Timeout while waiting for addon Cloudflared to start, took more then 120 seconds
```

I would still suggest not to pursue the topic at the moment.

PR #499, opened 2023-10-12
Update validHostnames to also allow Umlauts in Domain Name, execpt TLD

### LINKED ISSUE #496: Umlauts in the domain
### The problem

Hello,

I would like to use my existing domain for the addon, but unfortunately it is not possible because my domain contains umlauts.

The problem is quite easy to identify, the regex (validHostnameRegex) in cloudflared-config.sh prevents the use of umlauts, can this possibly be adjusted?

Best regards

4.2.7

_No response_

Home Assistant OS

### Add-on YAML Configuration

_No response_

_No response_

Specifies a domain with umlauts (ü, ö, ä) in the configuration for External hostnames. 

### Additional information

_No response_

brenner-tobias:
# Proposed Changes

Update the Regex for allowed Domain Names to include Umlauts.

Some considerations from my change:

1. In the part "left" of the TLD (so before the last .) I switched to [Unicode Categories](https://www.regular-expressions.info/unicode.html) for any lowercase letter `\p{Ll}` or decimal digit number `\p{Nd}`. I also changed the Quantifier before the TLD to a `+` to check for at least one "." and something before that in the domain name.
2. For the TLD itself, I did not make these changes, since there are from my knowledge no TLDs with Umlaute.

## Related Issues

closes #496

### DIFF
--- a/cloudflared/rootfs/etc/s6-overlay/scripts/cloudflared-config.sh
+++ b/cloudflared/rootfs/etc/s6-overlay/scripts/cloudflared-config.sh
@@ -14,7 +14,7 @@ checkConfig() {
     bashio::log.trace "${FUNCNAME[0]}"
     bashio::log.info "Checking add-on config..."
 
-    local validHostnameRegex="^(([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])\.)*([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])$"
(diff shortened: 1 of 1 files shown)


brenner-tobias:
I did test the regex with https://regex101.com/ but cannot test it right now in HA. @elcajon can you by any chance do that to make sure?

elcajon:
image
looks got to me 😉

elcajon (approved):

brenner-tobias:
Thanks a lot @elcajon

lkwr:
@brenner-tobias this merge request should be reverted because it prevent the addon to start with valid hostnames (see #500, #501, #502).

Also consider that there are tld domains like [`.中国`](https://en.wikipedia.org/wiki/.cn#Internationalized_domain_names_with_Chinese_characters), which contains special unicode characters as well.

Why don't leave the old regex and simply use [IDN](https://en.wikipedia.org/wiki/Internationalized_domain_name) format for domains with special unicode characters (which is already covered by the old regex)?

PR #505, opened 2023-10-13
Accept Umlauts & regular latin characters

### LINKED ISSUE #500: Addon wont start  "is not a valid hostname"
### The problem

addon wont start since update to 4.2.8

4.2.8

4.2.7

Home Assistant OS

### Add-on YAML Configuration

```
external_hostname: xxxxx.xxxxx.trade
additional_hosts: []
```

```txt
[18:57:16] INFO: Checking add-on config...
[18:57:17] FATAL: 'xxxxx.xxxxx.trade' is not a valid hostname. Please make sure not to include the protocol (e.g. 'https://') nor the port (e.g. ':8123') and only use lowercase characters in the 'external_hostname'.
s6-rc: warning: unable to start service init-cloudflared-config: command exited 1
/run/s6/basedir/scripts/rc.init: warning: s6-rc failed to properly bring all the services up! Check your logs (in /run/uncaught-logs/current if you have in-container logging) for more information.
/run/s6/basedir/scripts/rc.init: fatal: stopping the container.
```

Start the addon

### Additional information

_No response_

skipishere:
# Proposed Changes

Makes the regex work as before and keeps the umlauts  change while simplifies the regex logic as \w is the same as [a-zA-Z0-9_]

## Related Issues
Fixes #500 

### DIFF
--- a/cloudflared/rootfs/etc/s6-overlay/scripts/cloudflared-config.sh
+++ b/cloudflared/rootfs/etc/s6-overlay/scripts/cloudflared-config.sh
@@ -14,7 +14,7 @@ checkConfig() {
     bashio::log.trace "${FUNCNAME[0]}"
     bashio::log.info "Checking add-on config..."
 
-    local validHostnameRegex="^(([\p{Ll}\p{Nd}]|[\p{Ll}\p{Nd}][\p{Ll}\p{Nd}\-]*[\p{Ll}\p{Nd}])\.)+([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])$"
+    local validHostnameRegex="^(([\p{Ll}\w]|[\p{Ll}\w][\p{Ll}\w\-]*[\p{Ll}\w])\.)*([\p{Ll}\w]|[\p{Ll}\w][\p{Ll}\w\-]*[\p{Ll}\w])$"
 
     # Check for minimum configuration options
     if bashio::config.is_empty 'external_hostname' && bashio::config.is_empty 'additional_hosts' &&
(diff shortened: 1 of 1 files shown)


flufftronix:
This isn't getting a match to `pöp.example.com` for me at [regexr.com](https://regexr.com/)

skipishere:
> This isn't getting a match to `pöp.example.com` for me at [regexr.com](https://regexr.com/)
> pöp.example.com


https://regex101.com/ has the option to use a Python regex engine, whereas on that site I can't see that as an option

flufftronix:
> > This isn't getting a match to `pöp.example.com` for me at [regexr.com](https://regexr.com/)
> > pöp.example.com
> 
> https://regex101.com/ has the option to use a Python regex engine, whereas on that site I can't see that as an option

Gotcha. So the shell in this container uses Python for regex?

skipishere:
> > > This isn't getting a match to `pöp.example.com` for me at [regexr.com](https://regexr.com/)
> > > pöp.example.com
> > 
> > 
> > https://regex101.com/ has the option to use a Python regex engine, whereas on that site I can't see that as an option
> 
> Gotcha. So the shell in this container uses Python for regex?

Good question, I'm now not 100% sure. The latest change using `\w` should at least make it work as before (but ideally still work for the umlauts and other characters hopefully 🤞

Ideally there should be a test around this, but will admit my skills are more in C# & Regex (in .NET & JavaScript) and limited bash & python knowledge

brenner-tobias:
will be fixed in https://github.com/brenner-tobias/addon-cloudflared/pull/507, I will later on look for a more elegant solution

acshef:
> will be fixed in #507, I will later on look for a more elegant solution

@brenner-tobias may I suggest that umlauts (and other characters outside the ASCII character space) *don't* get this special treatment? Does Cloudflared (and by extension, Cloudflare/Cloudflare tunnel) work with punycode-encoded URLs? <https://en.wikipedia.org/wiki/Punycode>

Example:
`fürelise.com` --> `xn--frelise-n2a.com`

skipishere:
With further working, I think the final regex could be as simple as
`^((\w|\w[\w\-]*\w)\.)*(\w|\w[\w\-]*\w)$`

PR #507, opened 2023-10-13
Fix tegex with old approach plus Umlauts

brenner-tobias:
# Proposed Changes

Revert to the old Regex for valid hostnames and include ä,ö,ü,ß as well


elcajon (approved):

inverse:
Thanks for a quick fix :pray:

KimpiegamesYT1:
Still does not work for me with a .nl domain.

KimpiegamesYT1:
Still does not work for me with a .nl domain.

PR #522, opened 2023-11-10
Remove mapped folders

elcajon:
# Proposed Changes

There have been some changes on how to map folders (like config, share, ssl, backup...) into containers.
As we don't provide the option to set a specific data path any longer we can safely remove the mappings.

We can think about changing the default /data path to the new [add-on config structure](https://developers.home-assistant.io/docs/add-ons/configuration#add-on-advanced-options) so people will be able to get there tunnel.json and cert.pem easier. But I really don't know if it would be that useful. We also have a new source of errors if the files can be edited manually...

## Related Issues

### DIFF
--- a/cloudflared/config.yaml
+++ b/cloudflared/config.yaml
@@ -15,10 +15,6 @@ arch:
   - armhf
   - armv7
   - i386
-map:
-  - ssl:rw
-  - share:rw
-  - config:rw
 options:
   external_hostname: ""
   additional_hosts: []


brenner-tobias (approved):

brenner-tobias:
Like the clean up, thanks a lot for keeping this up-to-date.
I think it is fine for now not giving any more options, let's keep it as simple as possible as long as this is not requested

PR #538, opened 2023-12-10
Drop Support i386 & armhf

brenner-tobias:
# Proposed Changes

I would like to discuss if we should / want to / have to drop support for armhf & i386.
The support for those architectures was [dropped in the example addon](https://github.com/hassio-addons/addon-example/pull/153) as well as all major Home Assistant Community Add-ons.

For now, a base image is still available, though I am not sure which direction this is going and if we should already follow with not supporting these architectures anymore.

I would very much appreciate your thoughts.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -12,7 +12,7 @@
       "matchStringsStrategy": "any",
       "matchStrings": [
         "ARG BUILD_FROM=(?<depName>.*?):(?<currentValue>.*?)\\s+",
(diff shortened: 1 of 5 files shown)


elcajon:
I have also noticed that the architectures in most add-ons have been discontinued. I think we should make the change as soon as possible, the downloads of our image are also limited for the two architectures.

Unfortunately, there is not too much information about the background, I have only seen [this discussion](https://github.com/hassio-addons/addon-tailscale/discussions/292) about it

brenner-tobias:
Alright @elcajon , thanks for looking into it and giving your perspective. Let's move ahead and remove support.
Do you mind having a look at the PR?

elcajon (approved):

PR #573, opened 2024-01-17
Add run_parameters

brenner-tobias:
# Proposed Changes

Offer a string as option to add run parameters to cloudflared

## Related Issues

#572


brenner-tobias:
functionality works, documentation tbd.

brenner-tobias:
documentation done as well

cloudflared/config.yaml
```diff
@@ -23,8 +23,10 @@ schema:
...
       disableChunkedEncoding: bool?
   tunnel_name: str?
-  tunnel_token: str?
-  post_quantum: bool?
   catch_all_service: str?
   nginx_proxy_manager: bool?
+  tunnel_token: str?
+  post_quantum: bool?
+  run_parameters:
+    - str?
```
elcajon, line 31: what do you think about restricting the possible options to the one we want to allow to be set? Something like:

```yaml
run_parameters:
    - match(^(--region|--retries|--protocol)=.*$)?
```

elcajon (approved):
Please see comment. Otherwise LGTM but didn't test it.

cloudflared/config.yaml
brenner-tobias: Good idea, not sure though how to do it, since the value will be more then the parameter, e.g. "--region us" or "--protocol http2" or "--protocol quic" so we would need to specify all parameters plus there possible values, right?

brenner-tobias: ok I just realized my mistake, thaty why we have the =.*$ Let me work on that and test it. Will do tomorrow.

Thanks for the review and the suggestion

brenner-tobias: Should work now, ready for another review @elcajon

elcajon (approved):
LGTM 👍🏼

PR #608, opened 2024-03-05
Add --loglevel as valid run_parameter

### LINKED ISSUE #607: "Remote tunnel add-on setup" tunnels are always started with loglevel=info (perhaps "Local tunnel add-on setup" as well)
### The problem

The configuration option `log_level:` only applies to the add-on itself but I need `--loglevel=debug` applied to the tunnel startup for tunnel troubleshooting and I do not see any way of increasing the logging level.

5.1.4

_No response_

Home Assistant OS

### Add-on YAML Configuration

```yaml
external_hostname: ***.ca
additional_hosts:
  - hostname: ***.***.ca
    service: http://##.##.##.##:####
tunnel_name: homeassistant
tunnel_token: >-
  <redacted>
log_level: debug
nginx_proxy_manager: false
```
```


```txt
[03:15:30] INFO: Using Cloudflare Remote Management Tunnel
[03:15:30] INFO: All add-on configuration options except tunnel_token will be ignored.
[03:15:30] INFO: Connecting Cloudflare Tunnel...
[03:15:30] DEBUG: cloudflared tunnel --no-autoupdate --metrics=0.0.0.0:36500 --loglevel=info run --token=<redacted>
2024-03-05T11:15:30Z INF Starting tunnel tunnelID=<redacted>
2024-03-05T11:15:30Z INF Version 2024.2.1
2024-03-05T11:15:30Z INF GOOS: linux, GOVersion: go1.21.5, GoArch: arm64
2024-03-05T11:15:30Z INF Settings: map[loglevel:info metrics:0.0.0.0:36500 no-autoupdate:true token:*****]
```

Set `log_level: debug` and observe that remote managed tunnel is slways started with `--loglevel=info`

### Additional information

[Cloudflare Tunnel Run Parameters - ](https://developers.cloudflare.com/cloudflare-one/connections/connect-networks/configure-tunnels/tunnel-run-parameters/#logfile)
image

elcajon:
# Proposed Changes

SSIA

The `run_parameters` `--loglevel` option will override our previously set --loglevel option which should be fine as it has to be set on purpose. 

## Related Issues

resolves #607

### DIFF
--- a/cloudflared/DOCS.md
+++ b/cloudflared/DOCS.md
@@ -234,6 +234,7 @@ Valid parameters to add are:
 - --edge-ip-version
 - --grace-period
 - --logfile
+- --loglevel
 - --pidfile
 - --protocol
 - --region
@@ -255,6 +256,7 @@ as root. This path can be accessed, for example, via the VS-code add-on via
 run_parameters:
   - "--region=us"
   - "--protocol=http2"
+  - "--loglevel=debug"
 ```
 
 ### Option: `log_level`
--- a/cloudflared/config.yaml
+++ b/cloudflared/config.yaml
@@ -30,5 +30,5 @@ schema:
   tunnel_token: str?
   post_quantum: bool?
   run_parameters:
-    - match(^(--edge-bind-address|--edge-ip-version|--grace-period|--ha-connections|--logfile|--pidfile|--protocol|--region|--retries|--tag)=.*$)?
+    - match(^(--edge-bind-address|--edge-ip-version|--grace-period|--ha-connections|--logfile|--loglevel|--pidfile|--protocol|--region|--retries|--tag)=.*$)?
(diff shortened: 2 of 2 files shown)


brenner-tobias:
It will add the flag twice, right? So did you test if this will run without issues with the same flag twice and which flag value Cloudflared it would actually use?

elcajon:
Yes, I have tested this, both --loglevels are passed but only the latter is used by cloudflared.

brenner-tobias (approved):

brenner-tobias:
@elcajon I am thinking about removing the mapping of the log-level altogether, meaning that the log-level option of the add-on would only affect the actual add-on logs and not the Cloudflared logs, which is supposed to be changed manually.
Thoughts?

PR #609, opened 2024-03-06
Refactor log level logic

elcajon:
# Proposed Changes

As discussed the PR removes the connection from our `log_level` add-on option to the cloudflared tunnel log level.
`log_level` from now on sets the add-on log level only
`run-parameters` `--loglevel` can be used to change cloudflared tunnel severity (defaults to `info` if not set)

## Related Issues

### DIFF
--- a/cloudflared/rootfs/etc/s6-overlay/scripts/cloudflared-config.sh
+++ b/cloudflared/rootfs/etc/s6-overlay/scripts/cloudflared-config.sh
@@ -306,6 +306,30 @@ createDNS() {
     fi
 }
 
+# ------------------------------------------------------------------------------
+# Set Cloudflared log level
+# ------------------------------------------------------------------------------
+setCloudflaredLogLevel() {
+
+# Set cloudflared log to "info" as default
+CLOUDFLARED_LOG="info"
+
+# Check if user wishes to change log severity
+if bashio::config.has_value 'run_parameters' ; then
+    bashio::log.trace "bashio::config.has_value 'run_parameters'"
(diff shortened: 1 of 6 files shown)


cloudflared/rootfs/etc/s6-overlay/scripts/cloudflared-config.sh
```diff
@@ -306,34 +306,6 @@ createDNS() {
...
-        "error") log="error";;
-        "fatal") log="fatal";;
-    esac
-else
-    log="info"
-fi
-
-# Write log level to S6 environment
-printf "%s" "${log}" > /var/run/s6/container_environment/CLOUDFLARED_LOG
-CLOUDFLARED_LOG=${log}
```
brenner-tobias, line 332: We have to set this variable (CLOUDFLARED_LOG) somewhere else in the script from the run options of cloudflared if present in order for all the runs within this script (DNS creation, ...) to use the correct log-level.
Another option would be to remove the loglevel from the runs in here, which is probably ok but in the end not 100% correct.

brenner-tobias (changes requested):
See comment in line

cloudflared/rootfs/etc/s6-overlay/scripts/cloudflared-config.sh
elcajon: You are right, I first thought about removing it completely, but I implemented a logic to read the log level from the run_parameters, if set by the user. 
If you have any ideas on how to simplify the logic, please let me know.

PR #612, opened 2024-03-10
Remove remnants of armhf and i386 from workflows

felipecrs:
# Proposed Changes

Remove the remnants of `armhf` and `i386` from workflows from GitHub Actions workflows.

~Also, it simplifies the `linux/arm64/v8` platform to just `linux/arm64` to align with the examples from https://docs.docker.com/build/building/multi-platform. It works the same (let me know if you disagree).~

## Related Issues

- #538

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -146,14 +146,10 @@ jobs:
           echo "from=$from" >> $GITHUB_OUTPUT
           if [[ "${{ matrix.architecture}}" = "amd64" ]]; then
             echo "platform=linux/amd64" >> $GITHUB_OUTPUT
-          elif [[ "${{ matrix.architecture }}" = "i386" ]]; then
-            echo "platform=linux/386" >> $GITHUB_OUTPUT
-          elif [[ "${{ matrix.architecture }}" = "armhf" ]]; then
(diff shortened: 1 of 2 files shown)


elcajon (dismissed):
Nice catch, we forgot that when we were cleaning up. 

I don't know if the platform change will cause any problems, we took it over from the hassio-addons CIs back then. I would therefore trust your recommendation.

felipecrs:
Good to know. I'm sure this will not change anything, but I don't really care about it, and for the sake of keeping it aligned to the upstream, I'm going to revert it then.

https://github.com/hassio-addons/workflows/blob/0a1ecb299cbcd4a2cad810b8ca692dbc4831ff34/.github/workflows/addon-ci.yaml#L160

elcajon (approved):
👍

brenner-tobias:
Not sure if we should do this, the if statements do not really hurt, and might actually help, if for whatever reason we bring back the other Archs.
What do you think?

brenner-tobias:
Let's add it, you are right I dont see the Arch to come back, so thanks a lot @felipecrs

felipecrs:
Sorry the delay, but I would be ok with both options as well.

Issue #725, opened 2024-10-25
Add-On Metrics Not Exposed


xXAzazelXx:
### The problem

Hey Guys!

Great work on the add-on!

Unless I'm doing something wrong it looks like the add-on runs in bridge mode and does not map the metrics port so they are no accessible at all.

Is it possible to please have a look?

```
2024-10-25T13:19:41Z INF Starting metrics server on [::]:36500/metrics

sudo docker inspect ddcbe1fdee91 --format='{{.HostConfig.NetworkMode}}'
bridge
```

### What version of Cloudflared has the issue?

2024.10.0

### What was the last working version of Cloudflared?

2024.10.0

### What type of installation are you running?

Home Assistant OS

### Steps to reproduce the issue

locally from SSHing to HA

curl http://127.0.0.1:36500/metrics

curl: (7) Failed to connect to 127.0.0.1 port 36500 after 0 ms: Could not connect to serve

brenner-tobias:
Thanks for raising. You are right, currently the metrics server port is not exposed.
I just created a PR (#729) exposing the metrics server port.

brenner-tobias:
I just released [v5.2.0](https://github.com/brenner-tobias/addon-cloudflared/releases/tag/v5.2.0) which includes the exposed metrics port that can be configured in the configuration in the network section.
Let me know if there are any issues.

xXAzazelXx:
Thank you for having a look.
So is there anything i need to now enable in the configuration?


The documentation states that the metrics option is the default and I am still unable to see the exposed ports sorry
```

➜  ~ docker ps
CONTAINER ID   IMAGE                                                                 COMMAND                  CREATED          STATUS                  PORTS                                                                                                                                                                                                                                       NAMES
bddb71661d95   ghcr.io/brenner-tobias/cloudflared/amd64:5.2.0                        "/init"                  37 seconds ago   Up 36 seconds                           
```

brenner-tobias:
Yes, you have to go to the add-on configuration page, scroll down to the network section and add any port there, that you want to map to the metrics server in the container running on 36500/tcp
If you do not have any specific requirements port-wise I suggest to simply add 36500 there as well as I did in the following screenshot:

<img width="1085" alt="image" src="https://github.com/user-attachments/assets/ee0545bb-5d4d-4603-9702-c3d0980dc382">

xXAzazelXx:
i see, works perfectly now! 
thank you!

PR #729, opened 2024-10-27
Expose metrics port

brenner-tobias:
# Proposed Changes

Expose Metrics Server Port and add WebUI URL in Configuration

## Related Issues

#725

## Summary by CodeRabbit

- **New Features**
  - Introduced a metrics web interface for enhanced monitoring.
  - Updated port mapping for improved access to the metrics interface.

- **Configuration Improvements**
  - Reorganized development container settings for better clarity and structure.
  - Updated port configuration for development container to streamline access.
  - Added new network interface entries for metrics in multiple languages.


cloudflared/config.yaml
```diff
@@ -18,6 +19,10 @@ map:
 options:
   external_hostname: ""
   additional_hosts: []
+ports:
+  36500/tcp: 36500
+ports_description:
+  36500/tcp: Metrics Web interface
```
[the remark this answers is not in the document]
elcajon: I think we should add the description as [translation](https://developers.home-assistant.io/docs/add-ons/configuration#port-description-translations) to be consistent with the rest of our configuration descriptions.

I would suggest to set the port mapping to "null" to disable port forwarding per default.

elcajon (approved):
LGTM 👍🏼

brenner-tobias:
There is one issue with the way I implemented it: Since I added the `webui: http://[HOST]:[PORT:36500]/metrics` config, the link for the webui is visible in the add-on page. If you do not actively add the port though, clicking on the link loads a blank website `about:blank#blocked`

We do have three options:

1. Leave as it is right now
2. Remove the webui config, so the link is not shown and it is everyones own responsibility to configure the port and open the metrics link
3. Set the port to be present by default, so there are no issues for most users who do not disable it

I am leaning towards option 2, since this is a niche use case anyways, but I could also live with 3. 

@elcajon what do you think?

elcajon:
I would also tend towards option two, as otherwise the question would arise as to whether you need to forward the port on your router in order to be able to use the add-on and the tunnel at all...

cloudflared/config.yaml
```diff
@@ -7,6 +7,7 @@ description: "Use a Cloudflare Tunnel to remotely
 url: "https://github.com/brenner-tobias/addon-cloudflared/"
 codenotary: dev@brenner.tech
 init: false
+webui: http://[HOST]:[PORT:36500]/metrics
```
felipecrs, line 10: image

If your intention was to make a button show up here, I think it's missing ingress configuration too.

```yaml
ingress: true
ingress_port: 36500
ingress_stream: true # if websockets are needed, not sure
```

felipecrs: Nevermind, just saw https://github.com/brenner-tobias/addon-cloudflared/pull/730.

And I agree, I don't think the button should be added indeed for such interface.

BTW it worked fine:

image

brenner-tobias: Thanks, I think we are good with the current implementation, let me know if you have some suggestions for improvements.

PR #767, opened 2025-01-12
Add Nginx as a workaround to get log streaming working again

elcajon:
# Proposed Changes

This PR adds Nginx to this add-on and sets up a proxy to distinguish between HA and HA logs requests.

Thanks to @othorp  (https://github.com/home-assistant/supervisor/issues/5509#issuecomment-2585465570) for figuring out the correct proxy settings!

```
location ~ /api/hassio/.*/logs {
    proxy_cache off;
    proxy_buffering off;
    proxy_hide_header Content-Type;
    add_header Content-Type "text/event-stream";
}
```

**This PR will have no impact on currently running tunnels.**

It runs an nginx proxy on port 8321 (inside the Cloudflare add-on Docker) with the correct settings to get log streaming working again.

Can be testet with **remotely managed tunnels** by changing the host from `homeassistant:8123` to `localhost:8321`.

image

image

I would like to merge this without a disruptive change to continue testing until a workaround/solution is available.

If it seems to be stable after some first testing. 

We can think about changing the local managed tunnel uri to `localhost:8321` in the future as well, which would have some impact on the current following setup logic:

https://github.com/brenner-tobias/addon-cloudflared/blob/dc5aae2dc506543a8b575267f60f9077066cb969/cloudflared/rootfs/etc/s6-overlay/scripts/cloudflared-config.sh#L217-L232
 
For testing, clone the branch in /addons and then run the add-on locally with a tunnel token.

## Related Issues

https://github.com/brenner-tobias/addon-cloudflared/discussions/744

## Release Notes

- **New Features**
  - Added Nginx configuration for improved service proxying
  - Enhanced dependency management configuration

- **Bug Fixes**
  - Implemented workaround for live log streaming issue

- **Chores**
  - Updated Nginx version to 1.26.2-r4
  - Improved service management scripts for Nginx

- **Infrastructure**
  - Added new Renovate configuration for dependency tracking

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -24,6 +24,14 @@
       "datasourceTemplate": "github-releases",
       "depNameTemplate": "cloudflare/cloudflared",
       "versioningTemplate": "loose"
+    },
+    {
+      "customType": "regex",
+      "fileMatch": ["/rootfs/build.sh$"],
+      "matchStrings": [
+        "#\\s*renovate:\\s*datasource=(?<datasource>.*?) depName=(?<depName>.*?)( versioning=(?<versioning>.*?))?\\sARG .*?_version=\"(?<currentValue>.*)\"\\s"
+      ],
+      "versioningTemplate": "{{#if versioning}}{{{versioning}}}{{else}}semver{{/if}}"
     }
   ],
   "packageRules": [
--- a/cloudflared/rootfs/build.sh
+++ b/cloudflared/rootfs/build.sh
@@ -24,6 +24,13 @@ case $arch in
     ;;
 esac
 
+# Workaround for live log streaming issue
+# see https://github.com/brenner-tobias/addon-cloudflared/discussions/744
+
+# renovate: datasource=repology depName=debian_12/gnupg versioning=loose
+nginx_version="1.26.2-r4"
+apk add --no-cache nginx="${nginx_version}"
+
 # Download the cloudflared bin
 wget -O /usr/bin/cloudflared "https://github.com/cloudflare/cloudflared/releases/download/${cloudflaredRelease}/cloudflared-linux-${arch}"
 
--- a/cloudflared/rootfs/etc/nginx/template/nginx.conf.gtpl
+++ b/cloudflared/rootfs/etc/nginx/template/nginx.conf.gtpl
@@ -1,0 +1,50 @@
+daemon off;
+error_log stderr;
+pid /var/run/nginx.pid;
+
+events {
+    worker_connections 1024;
+}
+
+http {
+    map_hash_bucket_size 128;
+
+    map $http_upgrade $connection_upgrade {
+        default upgrade;
+        ''      close;
+    }
+
+    server_tokens off;
+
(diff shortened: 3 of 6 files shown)


brenner-tobias:
Thanks a lot! While testing one first thought: If we want to keep this as an optional workaround and expect a fix outside of this add-on at some point, would it not be sufficient to document a solution using the NGINX HA add-on and the tunnel with a catch-all rule on our end?

elcajon:
I assume so, but without further Nginx configuration this would not work via the official nginx add-on. I had already tested some configurations locally with Nginx and Caddy. It's not that easy to get it to work if you're not familiar with proxies.

I'm not so happy with this solution either, because I don't think it should be solved in the add-on, but I think for locally managed tunnels we could easily switch back if there is a solution upstream in the future.

brenner-tobias:
I tested it locally and it looks good to me.
Now I think we have to ask ourselves if we want to include this or if we just live with the limited functionality for now.

Have we identified any additional things (next to live logging) that are not working? If it is only that, I could live with the limited functionality, especially since the logs can be accessed while in the local network. 

On the other hand, I am also fine with integrating this workaround for anyone interested in fixing it themselves since it is not affecting anyone else. So if you want to include it, I am happy to approve the changes once the documentation is also updated with a comment and solution for the remote tunnels.

elcajon:
I see what you mean, and I don't feel comfortable adding something new that isn't really related to the core functionality of the add-on. I also don't want to extend that functionality in the future.

I don't think we have any other direct limitations apart from the live logs. But missing logs while debugging automations etc. seems to be crucial for me - also while away from home.

I honestly don't know if we should even mention it in the docs at this point. Anyone who comes across the problem and reads about it in the discussions could make the change. Otherwise, I would rather make it available quietly to not confuse the average user.

Another thought regarding the new live streaming from cameras (grpc) functionality, I really don't know if this will work with our current add-on configuration/solution, I don't have any devices to test this new functionality.

brenner-tobias:
fair point regarding the logging. Let's introduce it in a silent way and see what pops up in the issues and discussions.

brenner-tobias (approved):

elcajon:
Changing the categorization to bugfix to avoid issuing a minor version change

elcajon:
None that I would be aware of, maybe we should think about how to extend the function for the local managed tunnel

felipecrs:
I never tried the remotely managed tunnel because the local one always worked great for me.

But about this workaround with nginx, maybe we can consider using it by default? I haven't checked the implementation yet too, but I suppose the nginx process is always running anyway, and I also suppose there's no downside of using it.

elcajon:
I think we could do that, maybe we some edge version tests before releasing to stable. However, I don't expect there to be any problems, it has been running flawlessly for me with the remote managed tunnel since the beginning.

@brenner-tobias What do you think?

brenner-tobias:
yeah let's go for it. Maybe we introduce a flag, which is true by default, to deactivate the workaround if needed?

PR #781, opened 2025-01-31
Increase nginx proxy_read_timeout

elcajon:
# Proposed Changes

https://github.com/brenner-tobias/addon-cloudflared/discussions/744#discussioncomment-11977794
Increase timeout to 1d to prevent HA from duplicating latest log entries.

## Related Issues

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved proxy connection timeout for API log streaming, allowing longer-running log follow operations without interruption.

### DIFF
--- a/cloudflared/rootfs/etc/nginx/template/nginx.conf.gtpl
+++ b/cloudflared/rootfs/etc/nginx/template/nginx.conf.gtpl
@@ -40,6 +40,7 @@ http {
             proxy_set_header Upgrade $http_upgrade;
             proxy_set_header Connection $connection_upgrade;
             proxy_set_header X-Forwarded-Host $http_host;
+            proxy_read_timeout 1d;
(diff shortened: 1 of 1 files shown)


brenner-tobias (approved):

othorp:
I'm not sure this PR is effective to fix proxy timeout end-to-end.  proxy_read_timeout for Cloudflare itself is 100s: https://developers.cloudflare.com/fundamentals/reference/connection-limits/

Without being on an enterprise plan, I haven't found a way to override this.  I see timeouts for live-logs at 1.7min in the developer console, which aligns with the above.

elcajon:
I'm afraid you're right, I only tested for a little over a minute but not >100 seconds. I guess we then are only fixing a problem for customers who have subscribed to the enterprise plan...

PR #809, opened 2025-03-19
⬆️ Update nginx to 1.26.3-r0

elcajon:
# Proposed Changes

Current edge version isn't working anymore because of outdated nginx version. 

image

## Related Issues


brenner-tobias (approved):

brenner-tobias:
@elcajon thanks for this. I am afraid that this will also happen in the stable version if someone builds it, what do you think?
If so, I suggest to release the new version asap.

elcajon:
Good point, I will release a new version. I will also create a PR for this later today to move the apk logic from build.sh to the Dockerfile, just to make sure every building process checks the installed version instead of using the Docker cache.

PR #870, opened 2025-07-05
Avoid waiting for Home Assistant before starting add-on

felipecrs:
# Proposed Changes

I believe it doesn't make sense to wait for Home Assistant to start before starting this add-on. Especially considering it can be used to access other services beyond just Home Assistant.

Also, in case Home Assistant fails to start, this add-on shouldn't be blocked from starting.

https://developers.home-assistant.io/docs/add-ons/configuration/#optional-configuration-options

## Related Issues

This PR is a copy of https://github.com/hassio-addons/addon-nginx-proxy-manager/pull/613, which I thought made total sense.

## Summary by CodeRabbit

* **Chores**
  * Updated Cloudflared addon configuration with a new startup setting.

### DIFF
--- a/cloudflared/config.yaml
+++ b/cloudflared/config.yaml
@@ -6,6 +6,7 @@ description: "Use a Cloudflare Tunnel to remotely
   connect to Home Assistant without opening any ports"
 url: "https://github.com/brenner-tobias/addon-cloudflared/"
 codenotary: dev@brenner.tech
+startup: services
 init: false
 hassio_api: true
 hassio_role: homeassistant


felipecrs:
The only important API we call is to decide whether to use HTTPS or not to connect to Home Assistant:

https://github.com/brenner-tobias/addon-cloudflared/blob/c01f7e54bb0c31d4ca0465fcade372c1469013d1/cloudflared/rootfs/etc/s6-overlay/scripts/cloudflared-config.sh#L218

We could work around that by reading from HA's `configuration.yaml`.

elcajon (commented):
Makes sense 👍🏻

brenner-tobias:
I like the suggestion, thanks a lot @felipecrs.
Feel free to decide if we want to include this in this separate PR or add it to #843, I am fine with both. Since we have been running with this startup configuration for quite some time, there is no need to rush.

felipecrs:
Right. I'll probably keep this one separate, and will complete it eventually.

cloudflared/config.yaml
```diff
@@ -6,6 +6,7 @@ description: "Use a Cloudflare Tunnel to remotely
   connect to Home Assistant without opening any ports"
 url: "https://github.com/brenner-tobias/addon-cloudflared/"
 codenotary: dev@brenner.tech
+startup: services
 init: false
 hassio_api: true
 hassio_role: homeassistant
```
felipecrs, line 12: Will no longer be needed once this PR is complete.

```suggestion
```

brenner-tobias (commented):
Agreed, as long as we find another elegant way to check the configuration.yaml for an SSL configuration.
Another approach could be to default to http as long as HA is not running and check the config once it is up, but I am also happy with another approach

felipecrs:
This is working for me, what do you guys think?

felipecrs:
> Agreed, as long as we find another elegant way to check the configuration.yaml for an SSL configuration. Another approach could be to default to http as long as HA is not running and check the config once it is up, but I am also happy with another approach

@brenner-tobias I implemented what I consider to be "elegant". Please take a look. I think it makes sense and is nice, but please don't hesitate to disagree.

cloudflared/rootfs/etc/s6-overlay/s6-rc.d/nginx/run
```diff
@@ -19,9 +19,13 @@ if bashio::config.false 'use_builtin_proxy'; then
 fi
 
 bashio::log.debug "Merging options & variables for template"
+ha_port=$(yq '.http.server_port // 8123' /homeassistant/configuration.yaml)
+ha_ssl=$(yq '.http | has("ssl_certificate") and has("ssl_key")' /homeassistant/configuration.yaml)
```
brenner-tobias, line 23: This looks good to me, though I have some concerns: Since configuring HA is becoming more and more a "UI" play, I am wondering if checking the configuration.yaml will be sufficient in the future.

Also, I just checked the HA setting and found a setting that might already be an issue (see screenshot), if not setting it to "Automatic".

What are your thoughts?

image

felipecrs: I implemented an additional safeguard in case the configuration file cannot be found. Please check https://github.com/brenner-tobias/addon-cloudflared/pull/870/commits/c0c76516366131b274a5c6925b32509bfa53e162. Also, this code will be much cleaner after I port it to #843.

Home Assistant HTTP cannot be configured through the UI: https://www.home-assistant.io/integrations/http

That setting you mention does not alter Home Assistant HTTP settings, it's merely for figuring out the correct local address for use with the Companion app, for example.

And even if Home Assistant migrates someday the HTTP settings to UI, which I find unlikely, we will probably still be able to parse this setting from `/config/.storage`.

If that happens, we'll need to adapt our code, but because of the safeguards, the default port and no SSL will be assumed, which will probably be good enough to keep the add-on functioning for like 99% of the users.

brenner-tobias (approved):
Thanks for the clarification regarding the setting and your thoughts, as well as the safeguard.
I agree that with this set-up and no current possibility in HA, the benefits outweight the potential future fix we have to find, so I am happy with the implementation and merging it.
Thanks a lot for your awesome work for this add-on!

felipecrs:
Thanks for reviewing!

I added one more safeguard... against a broken YAML syntax.

elcajon:
I don't know if people still do this today, but what about those who organize their configurations in separate files?

image

brenner-tobias:
fair point @elcajon. I was thinking about how yq handles thos !include statements, but the only thing I found was [this question](https://github.com/mikefarah/yq/discussions/1906), so I guess it would be a manual scripting step for us to properly read through the include statements and check all the files, which I think is not worth the hassle.

So I suggest that we add a comment in the documentation (I would put it in the area [configuration.yaml](https://github.com/brenner-tobias/addon-cloudflared/blob/main/cloudflared/DOCS.md#configurationyaml) where the user has to change the file and the http integration anyways) and that's it.

elcajon (approved):

felipecrs:
> In that case, I suppose I'm the exception. 😉

lol, sorry! And with your screenshot, I wonder if I should also support `!secret`...

felipecrs:
> Seeing your screenshot, I wonder if I should also support `!secret`...

Probably not needed because:

1. I don't think anyone would use `!secret` for `server_port`
2. We don't need to read the actual value of `ssl_certificate` and `ssl_key`, we just need to check if these keys are set, meaning it will work even though `!secret` does not expand.

elcajon:
> Probably not needed because:
>
>I don't think anyone would use !secret for server_port
We don't need to read the actual value of ssl_certificate and ssl_key, we just need to check if these keys are set, meaning it will work even though !secret won't expand.

I would agree with that.

> lol, sorry!

No problem! I stole the idea from [here](https://github.com/frenck/home-assistant-config). Feel free to blame him. 😉

brenner-tobias:
Awesome @felipecrs and @elcajon, thanks a lot!

PR #886, opened 2025-08-11
Document that http in configuration.yaml must no use !include or packages

felipecrs:
# Proposed Changes

I would love avoiding this, but I don't think it's worth it.

I still think the gains outweigh the drawbacks though.

## Related Issues

- https://github.com/brenner-tobias/addon-cloudflared/pull/870#issuecomment-3172850721

## Summary by CodeRabbit

- New Features
  - More reliable detection of Home Assistant HTTP settings (port/SSL) by reading configuration directly from configuration.yaml.

- Documentation
  - Added guidance: if you change the Home Assistant port or enable SSL, keep those HTTP settings directly in configuration.yaml. Using !include or packages can prevent the add-on from determining the correct internal URL.
  - Added supporting cross-links for HTTP integration and Home Assistant packages.

- Refactor
  - Simplified configuration parsing to read from configuration.yaml, reducing complexity and improving robustness of internal URL detection.

### DIFF
--- a/cloudflared/DOCS.md
+++ b/cloudflared/DOCS.md
@@ -327,6 +327,13 @@ http:
 might have to add another IP(range) here. Check your HA logs
 after attempting to connect to find the correct IP.**
 
+If you changed your Home Assistant port or enabled SSL in the [HTTP
+integration][http-integration] configuration, please make sure to keep such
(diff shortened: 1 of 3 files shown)


elcajon (approved):
Thanks 👍🏻

PR #913, opened 2025-09-29
Remove built-in NGINX (workaround for live logs issues)

felipecrs:
# Proposed Changes

This removes the built-in NGINX proxy that was added as a workaround for live logs issues as apparently it is no longer needed.

## Related Issues

- Refs https://github.com/brenner-tobias/addon-cloudflared/pull/843#issuecomment-3330832795

## Summary by CodeRabbit

* **Chores**
  * Removed the built-in Nginx proxy option and related runtime components.
  * Ingress now connects directly to Home Assistant (no internal proxy path).
  * Build/container simplified by removing nginx-related packages and scripts.

* **Documentation**
  * Removed UI option and all translations for the built-in proxy.
  * Updated docs to remove references to the removed proxy option.


felipecrs:
PS: I didn't test this, but code changes seem in order.

felipecrs:
@elcajon this will be a breaking change once pushed for users that were previously using localhost:8321, however I think it is ok.

The adjustment is simple after all, just change back to homeassistant:8123.

Besides, it was a undocumented workaround.

brenner-tobias (approved):
Looks good, thanks a lot!

PR #1016, opened 2026-05-24
Update YQ version to 4.49.2-r6 in Dockerfile

sir-Unknown:
because the live build depends on the pinned Alpine package revision, and yq-go=4.49.2-r5 is no longer available.

# Proposed Changes

- Bump `yq-go` from `4.49.2-r5` to `4.49.2-r6`.
- Restore the Docker build for Home Assistant add-on installations that build from source.
- This also fixes the downstream Caddy variant based on this add-on.

## Related Issues

- None


elcajon (approved):

elcajon:
Thank you 👍🏻

PR #1049, opened 2026-07-30
fix: read configured http port from http config

### LINKED ISSUE #1048: 2026.8.0 breaks ports
### The problem

Hi!

Home Assistant `2026.8.0` introduces customizable ports. New installations now default to port `80` (enabling URLs like `http://homeassistant.local without specifying a port), and existing installs can now edit their port in settings.

Currently, this add-on assumes the port is `8123`, which will break for new installations or users with custom ports.

All

None

Home Assistant OS

### App (Add-on) YAML Configuration

```yaml

```

```txt

```

1. Install HA beta (`2026.8.0b1`)
2. Change your port in Settings -> System -> Network

### Additional information

A valid fix would be to detect the installation’s port (if possible?) or just add a setting in the app config.

RonaldPhilipsen:
# Proposed Changes

This PR fixes compatibility with ha 2026.8.0 by reading the http config from the local storage.
if we fail to parse the file or read from storage we fall back to the default behaviour, which is using port 8123

## Related Issues

fixes https://github.com/homeassistant-apps/app-cloudflared/issues/1048

### DIFF
--- a/cloudflared/rootfs/etc/s6-overlay/s6-rc.d/prepare/run.sh
+++ b/cloudflared/rootfs/etc/s6-overlay/s6-rc.d/prepare/run.sh
@@ -87,14 +87,20 @@ validateConfigAndSetVars() {
     local ha_port="8123"
     local ha_ssl="false"
     if bashio::var.is_empty "${external_hostname}"; then
-      bashio::log.debug "No external_hostname configured, skipping check of Home Assistant port and SSL"
+        bashio::log.debug "No external_hostname configured, skipping check of Home Assistant port and SSL"
     elif [[ -f "${ha_storage_http}" ]]; then
-      local ha_port_from_storage
-      ha_port_from_storage=$(yq '.data.stable.server_port // empty' "${ha_storage_http}" 2>/dev/null || true)
-      if [[ -n "${ha_port_from_storage}" && "${ha_port_from_storage}" != "null" ]]; then
-        ha_port="${ha_port_from_storage}"
-        bashio::log.debug "Read Home Assistant port from ${ha_storage_http}: ${ha_port}"
-      fi
+        local ha_port_from_storage
+        local ha_ssl_from_storage
+        ha_port_from_storage=$(yq '.data.stable.server_port' "${ha_storage_http}" 2>/dev/null || true)
+        ha_ssl_from_storage=$(yq '.data.stable | (has("ssl_certificate") and has("ssl_key"))' "${ha_storage_http}" 2>/dev/null || true)
+        if [[ -n "${ha_port_from_storage}" && "${ha_port_from_storage}" != "null" ]]; then
+            ha_port="${ha_port_from_storage}"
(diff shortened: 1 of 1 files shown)


brenner-tobias:
Thanks a lot for proposing this change. I am wondering, if this would be easier by using $(bashio::core.port)

E.g.:
```
# Get port and SSL status via built-in bashio helpers
CORE_PORT=$(bashio::core.port)
IS_SSL=$(bashio::core.ssl)

if bashio::var.true "${IS_SSL}"; then
    PROTOCOL="https"
else
    PROTOCOL="http"
fi

# Construct your origin target for cloudflared
TARGET_URL="${PROTOCOL}://homeassistant:${CORE_PORT}"
```

elcajon:
> Thanks a lot for proposing this change. I am wondering, if this would be easier by using $(bashio::core.port)

We used to do that until we replaced it with the current solution. --> https://github.com/homeassistant-apps/app-cloudflared/pull/920

I think another reason was that this allowed us to remove the supervisor "admin" (or manager) permissions for the add-on/app.

brenner-tobias:
ah yeah fair enough, I totally forgot about that, sorry.
Can you please also clarify what is not working in the current implementation anymore? It looks like it also defaults to 8123, but reads the port already.

RonaldPhilipsen:
> ah yeah fair enough, I totally forgot about that, sorry. Can you please also clarify what is not working in the current implementation anymore? It looks like it also defaults to 8123, but reads the port already.

long story short, I upgraded to  2026.8.0b1 and this prompted to remove the http block from the config as its now configured using the gui

elcajon:
> Can you please also clarify what is not working in the current implementation anymore? It looks like it also defaults to 8123, but reads the port already.

Unfortunately, I can't reproduce this, but since the configuration in HA has been changed from YAML to the UI, I don't see any reason not to implement the fix. Have you tested the changes, @RonaldPhilipsen?

RonaldPhilipsen:
got a bit distracted earlier, the grabbing of the port works, just missing setting the SSL enabled, debugging that locally now

RonaldPhilipsen:
@elcajon got this running again , took a second to figure out how to develop this locally

```
[08:04:13] INFO: Validating app (add-on) configuration...
[08:04:13] DEBUG: external_hostname: hass-test.domain.tld
[08:04:13] DEBUG: Checking Home Assistant port and if SSL is used...
[08:04:13] DEBUG: Read Home Assistant port from /homeassistant/.storage/http: 443
[08:04:13] DEBUG: Read Home Assistant SSL from /homeassistant/.storage/http: true
[08:04:13] DEBUG: ha_port: 443
[08:04:13] DEBUG: ha_ssl: true
[08:04:13] DEBUG: ha_protocol: https
[08:04:13] DEBUG: ha_url: https://homeassistant:443
[08:04:13] DEBUG: tunnel_name: homeassistant
```

elcajon:
Looks good to me, thanks a lot! @brenner-tobias what do you think, I would suggest to merge and release asap.

brenner-tobias (approved):
Looks good, thanks a lot!

PR #1063, opened 2026-08-11
fix: fix regressions in 7.0.12-13

### LINKED ISSUE #1060: 7.0.12/7.0.13: SSL misdetected as off when .storage/http has pre-2026.8 v1 flat schema — has() on missing .data.stable returns false, suppressing configuration.yaml fallback (http:// origin to TLS port, 502)
### The problem

Since 7.0.12/7.0.13, the add-on builds the Home Assistant ingress origin as `http://homeassistant:443` on a system where HA serves **TLS** on port 443 (`ssl_certificate`/`ssl_key` set in `configuration.yaml`, Supervisor `/core/info` reports `ssl: true, port: 443`). Every origin connection is then reset and the external hostname returns **502** from the Cloudflare edge. 7.0.11 worked.

The interesting part is the *asymmetry*: the **port** is detected correctly (443, from `configuration.yaml`), but **ssl** is not — so this isn't simply "the YAML fallback doesn't run". I traced it to the `.storage/http` handling in `prepare/run.sh` ([`validateConfigAndSetVars`](https://github.com/homeassistant-apps/app-cloudflared/blob/main/cloudflared/rootfs/etc/s6-overlay/s6-rc.d/prepare/run.sh)):

```bash
ha_port_from_storage=$(yq '.data.stable.server_port' "${ha_storage_http}" 2>/dev/null || true)
ha_ssl_from_storage=$(yq '.data.stable | (has("ssl_certificate") and has("ssl_key"))' "${ha_storage_http}" 2>/dev/null || true)
```

These paths assume the **HA 2026.8+ storage schema** (`STORAGE_VERSION 2`, where the payload is nested under `data.stable` — see [`homeassistant/components/http/config.py`](https://github.com/home-assistant/core/blob/2026.8.0/homeassistant/components/http/config.py): `KEY_STABLE: Final = "stable"`).

On HA Core **≤ 2026.7.x**, `.storage/http` (when present) uses the **v1 flat schema** — the settings live directly under `.data`, there is no `stable` key:

```json
{
  "version": 1,
  "minor_version": 1,
  "key": "http",
  "data": {
    "server_port": 443,
    "ssl_certificate": "/ssl/fullchain.pem",
    "ssl_key": "/ssl/privkey.pem",
    "...": "..."
  }
}
```

Against that file, the two yq expressions behave *differently* (yq v4):

```console
$ yq '.data.stable.server_port' http
null
$ yq '.data.stable | (has("ssl_certificate") and has("ssl_key"))' http
false
$ yq '.data | (has("ssl_certificate") and has("ssl_key"))' http   # correct path for v1
true
```

- The **port** read yields the string `null`, so the `[[ -n ... && ... != "null" ]]` guard rejects it and the script falls through to `configuration.yaml` → port 443. ✔
- The **ssl** read yields `false` — because `has()` on a missing (null) node returns `false`, not `null`. That passes the guard, is treated as an authoritative "storage says no SSL", and the `configuration.yaml` fallback for ssl is **skipped** (that skip is exactly what #1057 introduced). Result: `ha_url=http://homeassistant:443`.

So any install that (a) has a `.storage/http` file in the pre-2026.8 flat schema and (b) runs HA with TLS enabled via `configuration.yaml` gets a broken `http://` origin on its TLS port.

7.0.13 (and 7.0.12 by inspection — same storage-read logic)

7.0.11

Home Assistant OS

### App (Add-on) YAML Configuration

```yaml
external_hostname: home.example.tld
additional_hosts: []
```

HA Core 2026.7.4, `configuration.yaml`:

```yaml
http:
  server_port: 443
  use_x_forwarded_for: true
  trusted_proxies:
    - 172.30.33.0/24
  ssl_certificate: /ssl/fullchain.pem
  ssl_key: /ssl/privkey.pem
```

`.storage/http` exists with `version: 1, minor_version: 1` and the flat `data` layout shown above (contents mirror the YAML).

```txt
cloudflared ERR lines showed originService=http://homeassistant:443 with
"Unable to reach the origin service ... connection reset" for every request
(origin is TLS-only on 443, so a plaintext HTTP connection is reset).
Origin itself healthy: curl -sk https://homeassistant:443/api/ from another
add-on container returns 401 as expected.
```

(Exact startup lines rotated away before I captured them; the yq reproduction above is deterministic and shows the misdetection directly.)

1. HA Core ≤ 2026.7.x with TLS configured in `configuration.yaml` (`server_port: 443`, `ssl_certificate`, `ssl_key`) and a `.storage/http` file in the v1 flat schema (settings directly under `data`, no `stable` key).
2. Install/upgrade the add-on to 7.0.12/7.0.13 with `external_hostname` set.
3. Restart the add-on (no boot race involved — reproduces with Core fully up).
4. Generated ingress origin is `http://homeassistant:443`; external hostname 502s.

### Additional information

Suggested fix — make the storage read schema-aware and make "missing" distinguishable from "false":

```bash
# v2 (2026.8+): data.stable.*   |   v1 (≤2026.7): data.*
if [[ $(yq '.version' "${ha_storage_http}") == "2" ]]; then
    base='.data.stable'
else
    base='.data'
fi
# only trust storage ssl if the base node actually exists:
ha_ssl_from_storage=$(yq "${base} | select(type == \"!!map\") | (has(\"ssl_certificate\") and has(\"ssl_key\"))" "${ha_storage_http}" 2>/dev/null || true)
```

i.e. guard `has()` behind an existence check so a missing node yields empty/`null` (triggering the `configuration.yaml` fallback) instead of a false-authoritative `false`.

Related: #1048, #1053, #1055, #1057 (the change that skips the YAML fallback when storage "has" a value), and possibly the root cause of the origin-unreachable report in #1059.

Workaround that restores service on 7.0.13 (bypasses autodetection entirely): drop `external_hostname` and use an explicit `additional_hosts` entry:

```yaml
additional_hosts:
  - hostname: home.example.tld
    service: https://homeassistant:443
```

RonaldPhilipsen:
# Proposed Changes

Fix regression on pre-2026.8 versions.
Handle unknown storage versions with a warning.
Add a simple bash test suite to guard against any mishaps. 

## Related Issues

fixes #1060

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -119,6 +119,15 @@ jobs:
         with:
           prettier_options: --write **/*.{json,js,md,yaml}
 
+  test-shell:
+    name: Shell tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: ⤵️ Check out code from GitHub
+        uses: actions/checkout@v7
+      - name: 🚀 Run shell test suite
+        run: bash ./cloudflared/tests/test_config_and_set_vars.sh
+
   build:
     name: Build ${{ matrix.architecture }}
     needs:
@@ -130,6 +139,7 @@ jobs:
       - lint-prettier
       - lint-shellcheck
       - lint-yamllint
+      - test-shell
     runs-on: ubuntu-latest
     strategy:
       matrix:
--- a/cloudflared/rootfs/etc/s6-overlay/s6-rc.d/prepare/run.sh
+++ b/cloudflared/rootfs/etc/s6-overlay/s6-rc.d/prepare/run.sh
@@ -82,8 +82,8 @@ validateConfigAndSetVars() {
     data_path="/data"
 
     bashio::log.debug "Checking Home Assistant port and if SSL is used..."
-    local ha_config_file="/homeassistant/configuration.yaml"
-    local ha_storage_http="/homeassistant/.storage/http"
+    local ha_config_file="${HOMEASSISTANT_CONFIG_FILE:-/homeassistant/configuration.yaml}"
+    local ha_storage_http="${HOMEASSISTANT_STORAGE_HTTP:-/homeassistant/.storage/http}"
     local ha_port="8123"
     local ha_ssl="false"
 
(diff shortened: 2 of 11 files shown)


brenner-tobias (approved):
Thanks a lot!

PR #1083, opened 2026-09-06
Add optional MQTT status reporting for the Cloudflare Tunnel

vemboy200:
# Proposed Changes
This adds status reporting over mqtt, essentially a connection binary sensor. An opt in toggle is required to be on to enable the feature and when enabled the system auto discovers and connects to the ha's mqtt broker. Works well against my haos system with the mosquito broker.

A secondary pr (#1084) adds more features

Made with claude code

## Related Issues

nonw

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -126,7 +126,9 @@ jobs:
       - name: ⤵️ Check out code from GitHub
         uses: actions/checkout@v7
       - name: 🚀 Run shell test suite
-        run: bash ./cloudflared/tests/test_config_and_set_vars.sh
+        run: |
+          bash ./cloudflared/tests/test_config_and_set_vars.sh
+          bash ./cloudflared/tests/test_mqtt_status.sh
 
   build:
     name: Build ${{ matrix.architecture }}
--- a/cloudflared/DOCS.md
+++ b/cloudflared/DOCS.md
@@ -82,6 +82,7 @@ advanced config can be achieved using the remote tunnel setup.
 - [`post_quantum`](#option-post_quantum)
 - [`run_parameters`](#option-run_parameters)
 - [`log_level`](#option-log_level)
+- [`mqtt_status`](#option-mqtt_status)
 
 ### Overview: App configuration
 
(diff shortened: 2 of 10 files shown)


brenner-tobias:
Thanks a lot for your contribution.
I would like to start with a quick discussion weather this is necessary to be implemented into the add-on or if there are other possibilities to get the tunnel status on MQTT, since we want to keep the add-on as simple as possible.

@vemboy200: do you see any other possibilities?

@elcajon, @felipecrs, @lmagyar thoughts?

elcajon:
I'm still torn. I think we can retrieve the information from the metrics endpoint using a template or Bash sensor without adding an extension to the app. The extension also opens the door to further potential customizations. For example, what if someone isn't using the built-in MQTT app but another one? Then we'd need configuration options for additional MQTT servers, and so on.

For now, I'd rather vote against adding the extension to our system.

vemboy200:
> Thanks a lot for your contribution.
> I would like to start with a quick discussion weather this is necessary to be implemented into the add-on or if there are other possibilities to get the tunnel status on MQTT, since we want to keep the add-on as simple as possible.
> 
> @vemboy200: do you see any other possibilities?


My main use would to use this to see if the connection to my cloudflare tunnel is working. Home Assistant cloud has a similar entity like this called remote ui. If cloudflare or my internet is down then this is entity can tell me that remote access isn't working. It is opt in so it's not on by default.

Now for the concern related to mqtt brokers I could see the problem even though most users use mosquito.


> I think we can retrieve the information from the metrics endpoint using a template or Bash sensor without adding an extension to the app.

I think that might be a good idea instead of if I could do that. Although for the people wanting it may not be as plug-and-play. We could also make a custom integration, although that may be taking too far imo. Your thoughts on which one may be better?

vemboy200:
Ok claude pointed out (and i found out) that theres an api on port 36500, which i never knew until now because theres no documentation, should i just close this and #1084, make my own seperate pr for docs, and maybe make a custom integration utilizing this api?

elcajon:
From my perspective, that would at least be the fastest way to implement the requirement. I don't think we'll be implementing a solution like the one proposed in this PR anytime soon without further discussion.

felipecrs:
I believe you could use an integration like https://github.com/txxa/hass-openmetrics to scrape metrics from this endpoint.

lmagyar:
Just a note: you can force feed statuses to HA core directly, without any integration, just calling the API.

Docs says "You can create any state that you want, it does not have to be backed by an entity in Home Assistant." https://developers.home-assistant.io/docs/api/rest/

An example: https://github.com/lmagyar/homeassistant-addon-mariadb-inmemory/blob/master/mariadb/rootfs/usr/bin/report-sensors, and the API curl: https://github.com/lmagyar/homeassistant-addon-mariadb-inmemory/blob/master/mariadb/rootfs/usr/lib/api.core.sh

